PR #10, opened 2024-08-20
Komet changes for inclusion set. 

kec:
Changes for inclusion set. Also modified project pom by sorting properties, and adding javafx-maven-plugin and version. This may have been necessary secondary to transitive dependencies not having versions nailed down.


framework/src/main/java/dev/ikm/komet/framework/panel/axiom/ClauseView.java
```diff
@@ -277,6 +278,20 @@ private void setupForDefinitionRoot() {
 
     }
 
+    private void setupForInclusionSet() {
+        rootBorderPane.getStyleClass()
+                .add(StyleClasses.DEF_SUFFICIENT_SET.toString());
```
rbradley813, line 283: @kec, suggest implementing a style class for inclusion sets

kec: Agree... But before we can implement it, we need to design what it should look like? Circle again? Green or Blue? Etc. 

Can you add a backlog item for designing the CSS for axioms? 

For the moment, the current CSS is sufficient for Sept. I debated about cloning the DEF_SUFFICIENT_SET to get started (with the idea that we would revise it later), but it just touches more things (we would then have to update the CSS files...), and I was trying to keep the changes limited so we could get the capability to Eric...

rbradley813 (approved):

PR #13, opened 2024-08-20
Feature/iia 762 populate all the dropdowns for fqn

swaroopsalvi:


kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
```diff
@@ -97,32 +96,36 @@ public Set<ConceptEntity> findAllStatuses(ViewProperties viewProperties) {
...
     public Set<ConceptEntity> findAllCaseSignificants(ViewProperties viewProperties) {
-
-
-        //FIXME after connect-a-thon put this query back
-//        IntIdSet caseSenseDescendents = viewProperties.calculator().descendentsOf(TinkarTerm.DESCRIPTION_CASE_SIGNIFICANCE.nid());
-//        Set<ConceptEntity> allCaseDescendents = caseSenseDescendents.intStream()
-//                .mapToObj(caseNid -> (ConceptEntity) Entity.getFast(caseNid))
-//                .collect(Collectors.toSet());
-
-        return CASE_SIGNIFICANCE_OPTIONS;
```
dholubek, line 108: if we are able to refactor and not use CASE_SIGNIFICANCE_OPTIONS, then we can remove it from DataModelHelper

dholubek (commented):
looks good, just one suggestion about removing CASE_SIGNIFICANCE_OPTIONS from DataModelHelper

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
```diff
@@ -97,32 +96,36 @@ public Set<ConceptEntity> findAllStatuses(ViewProperties viewProperties) {
...
+        IntIdSet caseSenseDescendents = viewProperties.calculator().descendentsOf(TinkarTerm.DESCRIPTION_CASE_SIGNIFICANCE.nid());
```
carldea, line 99: Can we add a convenient function in `dev.ikm.komet.kview.mvvm.model.DataHelper` ?
```java
public static Set<ConceptEntity> fetchDescendentsOfConcept(ViewProperties viewProperties, PublicId publicId) {
   // return a set.
}

```

Now code can simply call this common function to get a set of concepts. This will remove the following common calls:

- Case significants
- Modules
- Path
- Or any similar use case.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
swaroopsalvi: Yes, the new starter data has the case significance options. I have uploaded the starter data in the story.
Also, Rohit said he will follow-up with data team for new database. May-connectathon data is not valid.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
swaroopsalvi: Done, please review. Do you want to call the helper method directly from the controller/View ? or via ViewModel ? I assumed via viewModel class, but then we can call it from view as well. The Helper resides in VM. Let me know if you need the call sequence to change and make it directly.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
dholubek: I was suggesting that we remove this from DataModelHelper... once you removed that return statement on line 108, then CASE_SIGNIFICANCE_OPTIONS is no longer referenced and can be removed.
https://github.com/ikmdev/komet/blob/ef95424ad6cad110509b8958a80bbfea436cd854/kview/src/main/java/dev/ikm/komet/kview/lidr/mvvm/model/DataModelHelper.java#L95

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DataViewModelHelper.java
carldea, line 1: Can you rename the file to DataModelHelper.java and move to the model package namespace?

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DataViewModelHelper.java
```diff
@@ -41,4 +47,12 @@ public class DataViewModelHelper {
             INSTANT_LITERAL,
             LONG
     ));
+
+    public static Set<ConceptEntity> fetchDescendentsOfConcept(ViewProperties viewProperties, ConceptFacade conceptFacade) {
```
carldea, line 51: Can the `conceptFacade` parameter be just the `PublicId` and not a `ConceptFacade`?

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
```diff
@@ -79,50 +78,32 @@ public DescrNameViewModel() {
         ;
     }
 
-    public Set<ConceptEntity> findAllLanguages(ViewProperties viewProperties) {
-        IntIdSet languageDescendents = viewProperties.calculator().descendentsOf(TinkarTerm.LANGUAGE.nid());
-        Set<ConceptEntity> allLangs = languageDescendents.intStream()
-                .mapToObj(langNid -> (ConceptEntity) Entity.getFast(langNid))
-                .collect(Collectors.toSet());
-        return allLangs;
+    public Set<ConceptEntity> findAllLanguages() {
```
carldea, line 81: Can we just remove this function?
The caller can call the `fetchDescendentsOfConcept()` directly.

swaroopsalvi:
There is no method to get descendants of using public id, so used
conceptFacade

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/properties/AddFullyQualifiedNameController.java
```diff
@@ -114,8 +114,8 @@ public void initialize() {
      * This copies form values into the ViewModel's property values. It does not save or validate.
      */
     private void copyUIToViewModelProperties() {
-        if (fqnViewModel != null) {
-            fqnViewModel.setPropertyValue(NAME_TEXT, fullyQualifiedNameTextField.getText())
+        if (descrNameViewModel != null) {
```
carldea, line 117: Can you remove the null check. Cognitive library injects the view model so it'll never be null. I know this was like that before, but I don't think we agreed to refactor all controllers. But since we are in here it can be easily fixed.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/DescrNameViewModel.java
dholubek: if this needs to be reverted based on leaving concept alone then disregard this comment.

PR #14, opened 2024-08-21
Bugfix and Evrete version update

andbi:
1. Migration to the newer 4.0.2 version.
2. Fixed a bug with loading rules.


carldea (approved):
Looks good to me!

PR #16, opened 2024-08-22
Reducing Evrete cold start time by an order of magnitude.

andbi:
This request converts Evrete literal rule conditions into method predicates, eliminates the need for the Java compiler, and solves the "pencil problem" - the high cold start time of the rule engine.

With the current count of rules, the new build time is around **50ms**.
Previous timings (on my Mac):
Versions 4.x.x: **~600ms**
Versions 3.x.x: **~2500ms**


carldea (approved):
Excellent work! Looks good to me.

PR #19, opened 2024-08-23
Bugfix/JPRO-18 Incorrect Data Source on App Startup

besidev:
The new implementation allows for different data services to be added or reordered in the `controllerOptions` list without breaking the default selection logic. It also ensures that the choice box will have a selected item even if the preferred default option is unavailable.


carldea (approved):
Looks good to me! Thanks @besidev

PR #20, opened 2024-08-23
Bugfix/Correct `mark` image URL in SelectDataSource.fxml

besidev:


application/src/main/resources/dev/ikm/komet/app/SelectDataSource.fxml
```diff
@@ -13,7 +13,7 @@
             <items>
                 <ImageView fitHeight="26.0" fitWidth="103.0" pickOnBounds="true" preserveRatio="true">
                     <image>
-                        <Image url="@../images/mark.png"/>
+                        <Image url="@mark.png"/>
```
carldea, line 16: **Question:** Does the file `mark.png` display correctly? 

I thought `@` means relative path where this fxml file resides. 
Screenshot 2024-08-23 at 10 55 31 AM

I'm not saying this won't work just wanted to make sure it will continue to work as an installed application and as a jpro project. 

:-)

carldea: Excellent work! I didn't realize there was another image. Thank you. I will approve and merge.

carldea (approved):
Looks great! thank you for testing and a snapshot of the change.

PR #24, opened 2024-08-26
Updated Build Workflow to generate installers.

OmOmofonmwan:


carldea (approved):
Looks good to me! Really great work btw. 
I think in the near future we can enhance the jobs we could do the following:

-  Create hash files for folks who download installers to be able to verify if binaries were altered.
- Include release notes and bug fixes.

PR #29, opened 2024-08-28
Bugfix/iia 906 user cannot see import option in app

swaroopsalvi:


application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -831,6 +831,7 @@ private void generateMsWindowsMenu(BorderPane kometRoot, Stage stage) {
 
         MenuItem importMenuItem = new MenuItem("Import Dataset");
         importMenuItem.setOnAction(actionEvent -> doImportDataSet(stage));
+        fileMenu.getItems().add(importMenuItem);
```
carldea, line 834: Can we make sure on the Mac Menu works too?

carldea (approved):
Looks good to me.

PR #31, opened 2024-08-30
Bugfix/iia 915 remove membership not working

dholubek:
Fixes a bug for the membership check that drives the menu on the journal>concept's identicon

image


carldea (approved):
Looks good.

PR #32, opened 2024-08-30
bugfix/IIA-919 and bugfix/IIA-920

dholubek:
bugfix/IIA-919 and bugfix/IIA-920
fixes the file naming and extension
keeps the export button enabled once the form is filled out.

image


carldea (approved):
Looks good to me!

PR #40, opened 2024-09-10
bugfix/IIA-921 export date dropdown match figma

dholubek:
For Export Data set, the Time period drop down should match as per Figma style 

Also fixes the name export text field color to be black instead of light grey

image

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/export/ExportController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/export/ExportController.java
@@ -277,8 +277,11 @@ public class ExportController {
             fromDate = this.customFromEpochMillis == 0 ? transformStringInLocalDateTimeToEpochMillis(dateTimeFromLabel.getText()) : this.customFromEpochMillis;
             toDate = this.customToEpochMillis == 0 ? transformStringInLocalDateTimeToEpochMillis(dateTimeToLabel.getText()) : this.customToEpochMillis;
         }
-        String namePrefix = exportName.getText().equals("") ? "komet" : exportName.getText();
-        String initialFileName = "%s-%s".formatted(namePrefix, simpleDateFormat.format(new Date()));
+        // if the user enters a name then use that name, e.g. test.json or test.zip
+        // if the user does not enter a name, then default to komet-yyyyMMdd-HHmm.zip|.json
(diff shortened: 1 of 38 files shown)


kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
```diff
@@ -2382,30 +2382,103 @@ Timeline Year line segment (will overlay change circles)
...
-    -fx-font-weight: 500;   
-    -fx-background-color: -Grey-1;
-}
+    -fx-font-weight: 500;
 
-.blue-arrow-combo-box-grey11.combo-box-base {
-    -fx-border-radius: 4;
-    -fx-border-color: -Grey-5;
-    -fx-background-color: -Grey-1;
+    /*-width-size: 180px;*/
```
swaroopsalvi, line 2397: Can we remove -width-size: 180px; 
since its commented?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/export/ExportController.java
```diff
@@ -242,6 +244,7 @@ private PopOver createPopover(UUID exportTopic, final int rangeType, Consumer<Lo
     public void handleCurrentDateTimeExport() {
         // Responsible for showing/hiding custom date range controls
         timePeriodComboBox.valueProperty().addListener((obs, oldVal, newVal) -> {
+            //FIXME
```
swaroopsalvi, line 247: Can you make this into a ToDo comment and add more details on what needs to be fixed ?

carldea (approved):
Looks good to me! Thanks.

swaroopsalvi (commented):
@dholubek , just have 2 small review comments.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/export/ExportController.java
dholubek: this comment is old and will remove it.

kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
dholubek: yes

PR #46, opened 2024-09-13
IIA-925 Updated stamp view models to use enums as property names

carldea:
Cognitive (MVVM) framework now allows property names to be enums which enables code to be more concise instead of declaring `public static final String PROPERTY_NAME = "some property name";`

In this branch the is the first introduction to using the new feature by using an existing enum in tinkar-core called `StampFields`.

Screenshot 2024-09-13 at 12 49 14 PM
For example: When defining ViewModels you can add properties like so:
Screenshot 2024-09-13 at 12 51 56 PM

As you will notice the code above uses static imports which will make properties less verbose.

At a later time we can begin to refactor and reduce a lot of boiler plate code that represents Form Fields in UIs.


dholubek (approved):
looks good

PR #47, opened 2024-09-13
Update owlapi version

emays:
Fixes issue with jpms deps


carldea (approved):
Looks good to me. We will test before we merge.

PR #50, opened 2024-09-16
Bugfix/Fix `kview.css` stylesheet refreshing when its updated dynamically

besidev:


carldea (approved):
Looks good to me. Thanks for cleaning this up. I believe @FlorianKirmaier mentioned a library that provides this refresh ability.

besidev:
That’s right, Carl! There’s a library called [CSSFX](https://mvnrepository.com/artifact/fr.brouillard.oss/cssfx/11.5.1), but it needs to be modularized before we can use it with Komet. We’ll provide a JPMS-compliant version in a separate PR.

PR #57, opened 2024-09-18
IIA-953 bugfix to remove the CMD key detection on the MacOS platform.…

carldea:
For scalable desktop on the journal view of Komet a user performs panning using the Command key and mouse drag operation. However, on MacOS many keys use the CMD key for instance CMD + SHIFT + 4 is used to do a screen shot. 
If there are any popup controls such as combobox, hover tooltips, Context menus, or menus you can not screen shot them b/c once you press the CMD key the popup disappears.

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
@@ -302,9 +302,9 @@ public class JournalController {
 
     private void setupScalableDesktop() {
         journalBorderPane.addEventHandler(KeyEvent.KEY_PRESSED, keyEvent -> {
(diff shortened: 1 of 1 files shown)


dholubek (approved):
looks good

carldea:
cc/ @besidev 

Besmir, 
I removed the command key check (keypress detection). The command key on macos affects the ability to take screen shots whenever there is a popup control such as context menus or tooltips. Let me know if there are any issue when using a browser (JPro).

Thanks for the contribution.
Carl 
:-)

besidev:
Thank you for the update Carl. We have tested the changes on the browser using JPro, and I can confirm that there are no issues. Everything is working smoothly without the command key check (which was not working since it interact with the browser functionalities depending on the platform).

Thanks again for the change,
Besmir

PR #61, opened 2024-09-20
Bugfix/JPRO-45 Unable to Run Reasoner when running Komet with JPro

besidev:
**Current behavior:**
When user runs the reasoner, nothing happens and the app freezes. This does run on the desktop app.

**Expected behavior:**
When user runs the reasoner, the following should happen:

- Reasoning process is initiated
- Activity tab shows progress of reasoning progress
- Changes from reasoning should be reflected in concept details

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
@@ -399,7 +399,7 @@ public class JournalController {
         Subscriber<ProgressEvent> progressPopupSubscriber = evt -> {
             // if summon event type, load stuff and reference task to progress popup
             if (evt.getEventType() == SUMMON) {
-                progressToggleButton.setVisible(true);
+                Platform.runLater(() -> progressToggleButton.setVisible(true));
                 Task<Void> task = evt.getTask();
                 JFXNode<Pane, ProgressController> progressJFXNode = createProgressBox(task, evt.getCancelButtonText());
                 ProgressController progressController = progressJFXNode.controller();
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
@@ -412,10 +412,7 @@ public class JournalController {
                     progressController.cleanup();
                 });
 
-                popOver.setOnHidden(windowEvent -> {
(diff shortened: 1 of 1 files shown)


besidev:
cc/@carldea

Carl,
Just a quick question here: since the entire `progressPopupSubscriber` code runs on the event bus working thread (Tinkar-Q-work-thread), and all of the code within the `if` block is mostly JavaFX-related, shouldn't we **wrap** it all inside a single `Platfrom.runLater` call to ensure it executes on the `JavaFX Application Thread`. What are your thoughts about this?

Thank you,
Besmir

carldea:
@besidev Besmir,
I think you might be correct (let's do it 👍🏼 ). Please feel free to update the code to be cleaner and more readable, but most importantly not block the UI thread. While what you did is better it does seem small sections that need it and some sections that don't. Making the whole block is fine as long as there isn't an expensive call inside the JavaFX controller code's (`initialize()` method). Of course we would test the many scenarios: Reasoner, Export Change set, Export FHIR.

In the beginning I was debating on a way for the event bus to wrap code as a Consumer to `Platform.runLater()`, but I wanted to keep things agnostic (pure Java) at first. In the future we could write a version of an event bus to make it easier for the Subscriber developer to perform work in the correct thread context (Application thread).

**Some additional notes:**
The Tinkar thread pool is used for long running callables (not on the JavaFX thread). The Progress Indicator in the Journal view is new and will continue to run on these Executors. (I recently created it for developers to spawn tasks that would have an opportunity to update progress). 

The Komet framework APIs provide a pure Java facility to decouple progress from JavaFX progress called `dev.ikm.tinkar.common.service.TrackingCallable` . Classic Komet is able to display progress or activity on running tasks when run using the Tinkar thread pool.  It's often hard to take different ways (past developer's) to run long running tasks such as Runnables, Futures, and Callables. I believe there is code we need to revisit to support cancelling long running tasks.

besidev:
Hello Carl,

Thank you for your feedback! I have proceeded to update the code by wrapping the JavaFX-related code inside a single `Platform.runLater` method call. You’re right, we should be cautious with the JavaFX controller’s `initialize()` method to avoid any expensive calls.

I appreciate the insights on the event bus and threading. Integrating a version that allows subscribers to perform work on the correct thread context sounds like a valuable future enhancement.

Thanks again for the guidance!

carldea:
Looks good to me! I am going to approve and merge the PR.

carldea (approved):
Looks good to me. Thank you!

PR #64, opened 2024-09-23
bugfix/IIA-885 For creating the new concept "Add Fully Qualified" button should be replaced as "Add Fully Qualified Name"

dholubek:
update menu item, fix source of events

image


carldea (approved):
Looks good to me.

PR #63, opened 2024-09-21
Bugfix/JPRO-50 Komet Bug: LIDR Record issue

besidev:
**Current behavior**: 
When user tries to create a LIDR record, and adding the device and click done, the entire LIDR record vanishes and the app freezes.

**Expected behavior**: 
User should be able to create the LIDR record. Once LIDR record is created, user should be able to find it in hierarchy navigation panel under Model concepts &gt; Device.

PR #67, opened 2024-09-24
IIA-954 bugfix lidr window disappears device pane when done button press

carldea:
Create Lidr window will shift view (jumps) to the right of the desktop workspace when ever a user clicks on the done button on the device selection screen. This also occurs when user is pressing done in the Analyte Group window.

Screenshot 2024-09-23 at 10 18 54 PM

Another window having the same issue. When the done button is pressed the bump out will close and jump to the scrollpane to the center or far right side instead of upper left screen.

Screenshot 2024-09-23 at 10 19 59 PM

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/lidr/mvvm/view/analyte/AnalyteGroupController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/lidr/mvvm/view/analyte/AnalyteGroupController.java
@@ -147,7 +147,7 @@ public class AnalyteGroupController implements BasicController {
         // order to save/update the device and manufacturer concepts
 
         clearView();
-        doneButton.disableProperty().bind(analyteGroupViewModel.getProperty(SAVE_BUTTON_STATE));
+        doneButton.disableProperty().bind(analyteGroupViewModel.getProperty(IS_INVALID));
 
         // setup drag n drop
         setupDragNDrop(analyteSearchStackPane, (publicId) -> {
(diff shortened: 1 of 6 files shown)


carldea:
@besidev Hello Besmir! 
FYI, based on your PR (https://github.com/ikmdev/komet/pull/63) I've cleaned up the code a bit and had taken a slightly different approach where I have validation code that gets triggered to update the `IS_INVALID` property on the view model. The **Done** button's **disable** property binds to the `IS_INVALID` property. 

I'm not exactly happy with this strategy, but it is much more predictable as a user performs data entry, clearing the form or canceling. 

I wasn't able to add you as a reviewer, but I'm tagging you here. Please review to compare against your PR. I'm still not sure why the issue is happening when the user presses the **Done button** and simple code that merely does the following: `doneButton.setDisable(true);`. 

By avoiding setting the property after clicking on the **Done button** and clearing the screen (and closing the bump out), it doesn't jump or disappears. When a subsequent display of the Device pane, the code will update the view model property `IS_INVALID` (to `true`) that will change the button's disable property state.

dholubek (approved):
looks good

besidev:
@carldea 
Hello Carl,

First, thank you for explaining why you took a slightly different approach in creating this PR. I also came to the same conclusion that the line `doneButton.setDisable(true);` was causing the issue, but as you, I’m still not entirely sure why it’s happening. After some more tests, it seems that when entering the `disabled` state, the button triggers a layout **recomputation** inside the parent `ScrollPane`, or more specifically within the `ScrollPaneSkin` default JavaFX implementation code. So far, the only solution we have is to avoid calling that line after pressing the `DONE` button, as already included in this fix, even though I'm not that happy either because the root cause is still unknown.

I really like your approach and the changes you’ve made based of using a boolean property from the view model to disable the button right before opening the Device pane for example, instead of using the `EvtBus` triggered by the `OPEN_PANEL` event type. It’s cleaner and seems to handle user interactions more predictably by utilizing the view model validation logic.

Looks very good to me 🙂!

carldea:
@besidev Besmir,
Glad we're thinking the same!

You explained it well wrt to the **recomputation** or (invalidation triggering a layout children). I even tried to set the button's `setManaged(false)`. I've tried many other things, except really digging in to the source code (`ScrollPaneSkin`). But, we will be tracking this issue closely and get a better understanding. I'll keep you posted.

PR #68, opened 2024-09-24
bugfix/IIA-939 pattern properties bumpout second toggle should be 'instances'

dholubek:
Second toggle for create Pattern will say "instances" instead of "hierarchy"

image


swaroopsalvi (approved):
Approved

PR #69, opened 2024-09-24
Feature/iia 778 pattern save one transaction

dholubek:
Saving a pattern in one transaction.  Not using the uncommitted timestamp.  NOTE: we will be upgrading Tinkar Composer from 1.0.0 to 1.2.0

before running unit test
image

after running unit test, we see the public Id of our pattern in the log
image

relaunch Komet to confirm that 1) it exists and 2) the public id matches
image

Pattern Other Name: [07d68747-fc9e-48e4-a66c-e130c7e372a6]

update... creating a pattern with a field definition:
image


application/src/test/java/dev/ikm/komet/app/test/PatternViewModelTest.java
```diff
@@ -143,68 +123,20 @@ public PublicId setPatternDefintion() {
         PublicId patternPublicId = PublicIds.newRandom();
         Pattern pattern = Pattern.make(patternPublicId);
 
+
         // the composer handles saving to an uncommitted stamp
         // it is un-committed until you say commit
         session.compose((PatternAssembler patternAssembler) -> patternAssembler
                 .pattern(pattern)
                 .meaning(patternMeaning)
                 .purpose(patternPurpose)
```
carldea, line 132: Should we add fields? 
Or is there a test with a larger Pattern record?

dholubek: there are field values see line 211/143 = old/new

dholubek: @carldea  (setting back to unresolve) there aren't fieldDefinitions in this unit test, but we can add them.  IIA-772 does have them

dholubek: added

carldea (approved):
looks good to me!

PR #74, opened 2024-09-30
Feature/iia 946 allow users to delete fields from the list of created fields

swaroopsalvi:


kview/src/main/java/dev/ikm/komet/kview/fxutils/MenuHelper.java
```diff
@@ -126,7 +126,7 @@ public static void fireContextMenuEvent(ActionEvent actionEvent, Side side, doub
      * This method takes list of menuItems in 2D array form.
      * The 2nd array can have the following items in sequence: public NAME = 0; ENABLED = 1; STYLES = 2; ACTION = 3; GRAPHIC = 4;
      * @param menuItems
-     * @return
+     * @return contextMenu
```
carldea, line 129: Can you elaborate with a sentence?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -427,6 +433,8 @@ private Node createFieldEntry(PatternField patternField, int fieldNum) {
     }
 
     private void showContextMenuPopUp(PatternField patternField, Node node, double x, double y) {
+        selectedPatternField = patternField;
+        selectedNodeField = node;
```
carldea, line 437: **Question:** why are we creating 2 global variables of a selected field?

My previous comment still applies. Based on the method name, it (`showContextMenuPopUp()`) is doing more than showing the context menu options. Lambdas (on click event listener) does allow you to assign global variables without being `final`. 

Having said this (assuming you have global variables) you have two choices:

1. Remove method and put code into the label's onclick handler
2. Change the name of the function to be `selectPatternField(PatternField patternField);` the onclick handler would call this method and another statement to call `contextMenu.show(label, x, y);`

kview/src/main/java/dev/ikm/komet/kview/fxutils/MenuHelper.java
swaroopsalvi: added more details on the returned value.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
swaroopsalvi: Thanks Carl, I have put the code in setOnMouseClicked and assigned the global variables.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -304,12 +307,17 @@ private void setupContextMenuMenuItems() {
...
                 {"Add Comment",  false, new String[]{"menu-item"}, null, createGraphics("comment-icon")},
-                {"Remove", true, new String[]{"menu-item"}, null, createGraphics("remove-icon")}
+                {"Remove", true, new String[]{"menu-item"}, (EventHandler<ActionEvent>) this::removeField, createGraphics("remove-icon")}
         };
 
         contextMenu = MenuHelper.getInstance().createContextMenuWithMenuItems(menuItems);
     }
 
+    private void removeField(ActionEvent actionEvent) {
+        fieldsTilePane.getChildren().remove(selectedNodeField);
```
carldea, line 317: Can we remove this line? This statement should be in the `FIELDS_COLLECTION` observable list change listener.

carldea: **Can we wire things up as we discussed related to the two phases?**

1. Subscriber (listens for delete event),
2. ObservableList (view model's list) add change listener to update UI  
While the ui is the `fieldsTilePane.getChildren()` it may or may not have its own change listeners. In this case it does namely updateField numbers (order) on labels.

There should be an Event subscriber that listens for fields being deleted this would be responsible for removing the field from the  `patternViewModel.getObservableList(FIELDS_COLLECTION).remove(patternField)`.

a change listener on that observable list above listens for a remove, if so, removes it from the `fieldsTilePane.getChildren().remove(selectedNodeField);`

dholubek (approved):
looks good

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -165,6 +165,9 @@ public class PatternDetailsController {
...
 
+    private Node selectedNodeField;  // this variable is used to select the node that was clicked on.
+    private PatternField selectedPatternField; //select the pattern field object for the node.
```
carldea, line 169: Can we just store the index (int) of the item in the List?

carldea (approved):
Things look good to me.

PR #75, opened 2024-10-04
Feature / JPRO-38 Prevent Concept Windows from going outside the Workspace

indritbeqiri:
As a User, I want to not lose concept windows I am working in by accidentally dragging them outside of the workspace.

**Acceptance Criteria**:

- All four edges of the workspace are blocked off so that the user is unable to drag concept windows beyond the edge of the workspace.
- The edge of a concept window should not go past the edge of the workspace.

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/fxutils/window/WindowSupport.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/fxutils/window/WindowSupport.java
@@ -90,30 +90,55 @@ public class WindowSupport {
     /**
      * Desktop area surface.
      */
-//    private final Pane desktopPane;
+    private final Pane desktopPane;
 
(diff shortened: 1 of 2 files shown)


carldea (approved):
@indritbeqiri  Nice work!

Glad you were able to clamp a window's position with the existing way the code currently looks like. 
:-)

**My thoughts out loud:**
1. Another solution could be to create a custom layout to manage child windows. 
2. In the **future** I was hoping that once we get the main features for scalable desktop, It would be nice to create a custom control or custom container as a separate jpms module.

indritbeqiri:
@carldea Thank you very much for the feedback!

The initial goal was to work within the current code structure without making major changes.

That said, it’s great to see that your thoughts are perfectly in line with our vision for the future UI architecture.
@indritbeqiri @besidev

PR #76, opened 2024-10-04
Feature/iia 771 implement functionality to edit field name

swaroopsalvi:


dholubek (approved):
looks good

kview/src/main/java/dev/ikm/komet/kview/events/pattern/PatternFieldsPanelEvent.java
```diff
@@ -23,6 +23,10 @@ public class PatternFieldsPanelEvent extends Evt {
 
     public static final EvtType<PatternFieldsPanelEvent> PATTERN_FIELDS = new EvtType<>(Evt.ANY, "PATTERN_FIELDS");
 
+    public static final EvtType<PatternFieldsPanelEvent> EDIT_FIELDS = new EvtType<>(Evt.ANY, "EDIT_FIELDS");
```
carldea, line 26: FYI,
Event design allows you to have a event type hierarchy assuming that is what you want to represent. Since I'm not familiar with the flow I'm assuming EDIT_FIELDS and ADD_FIELDS is a type of PATTERN_FIELDS. And if so, can be like the following:
```java
public static final EvtType<PatternFieldsPanelEvent> EDIT_FIELDS = new EvtType<>(PATTERN_FIELDS, "EDIT_FIELDS");
```
You may want to check with @dholubek

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
```diff
@@ -157,24 +164,35 @@ private void initialize() {
...
         displayNameProp.addListener(fieldsValidationListener);
         dataTypeProp.addListener(fieldsValidationListener);
-        purposeProp.addListener(fieldsValidationListener);
-        meaningProp.addListener(fieldsValidationListener);
+        purposeProp.addListener((obs, oldVal, newVal) -> {
+            if(newVal != null){
+                addPurposeToForm(purposeProp.getValue());
+            }
+            fieldsValidationListener.changed(obs, oldVal, newVal);
+        });
```
carldea, line 188: This should not be combined. It should be the following:
```java
purposeProp.addListener(fieldsValidationListener);
purposeProp.addListener((obs, oldVal, newVal) -> addPurposeToForm(newVal)); 
// the function should do the null check.
```
Same goes for meaning below.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/PatternViewModel.java
```diff
@@ -63,6 +63,9 @@ public class PatternViewModel extends FormViewModel {
 
     public static String MEANING_DATE_STR = "meaningDateStr";
 
+    // Used to load the values in the PatternField controller from PatternDetailsController.
+    public static String SELECTED_PATTERN_FIELD = "selectedPatternField";
```
carldea, line 67: I thought we discussed this in detail. Selected items for editing or removing does not need to be maintained. The observable list of Fields collection are indexed.

swaroopsalvi: @carldea , I think what we discussed was, said that we would not be creating an instance in the controller class and we are okay in passing the actual value around via ViewModel. 
     Hence did the variable creation. This is required when we edit a field and change its field position, for example we created 3 fields, and then we want to edit field 3 and change its field order to 1. In this scenario, we will have to retain either the selected field order or selected pattern field, I chose the later. And then when we come back to the details screen we can use it to remove and then add it to the desired position so that the order is changed. that is Field 3 becomes1, Field 1 becomes 2 and Field 2 becomes 3.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
swaroopsalvi: I will fix this

kview/src/main/java/dev/ikm/komet/kview/events/pattern/PatternFieldsPanelEvent.java
swaroopsalvi: Will check with Dough.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/PatternViewModel.java
swaroopsalvi: After talking with @dholubek he said we can discuss the use of the Axiomic API with you on this implementation.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -264,6 +265,10 @@ public void onChanged(Change<? extends DescrName> change) {
         patternFieldsPanelEventSubscriber = evt -> {
             PatternField patternField = evt.getPatternField();
             int fieldPosition = evt.getCurrentFieldOrder()-1;
+            if(evt.getEventType() == EDIT_FIELD){
+                // 1st remove it from list before adding the new entry
+                patternFieldList.remove(patternViewModel.getPropertyValue(SELECTED_PATTERN_FIELD));
```
carldea, line 270: This statement is not needed. The event's patternField is all you need. This subscriber will update the FIELD_COLLECTION (observable list). The FIELD_COLLECTION's listener will do the removing of the UI's VBox item. Remember there are two collections, they are in sync. If you remove and item (by index) it's the same remove. 
Assuming you are doing a remove and an insert(add).

kview/src/main/java/dev/ikm/komet/kview/events/pattern/ShowPatternFormInBumpOutEvent.java
```diff
@@ -40,6 +42,8 @@ public class ShowPatternFormInBumpOutEvent extends Evt {
     public static final EvtType<ShowPatternFormInBumpOutEvent> SHOW_EDIT_FQN = new EvtType<>(DESCRIPTION_NAME, "SHOW_EDIT_FQN_DESCRIPTION");
     public static final EvtType<ShowPatternFormInBumpOutEvent> SHOW_EDIT_OTHER_NAME = new EvtType<>(DESCRIPTION_NAME, "SHOW_EDIT_OTHER_NAME_DESCRIPTION");
     private final int totalFields;
+    private final PatternField patternField;
+    private final int fieldOrder;
```
carldea, line 46: Can we not have these global variables?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
swaroopsalvi: Hi Carl, I think that is exactly what the above code is doing. Removing it from the FIELDS_COLLECTION the listener of the FIELDS_COLLECTION is updating the UI.  There is no other place in this controller class code where we are removing the selected pattern. 
    The events patternField is the new value/instance that is add to the FILEDS_COLLECTION list and to the UI via its listener. It is not related or same as the to the SELECTED_PATTER_FIELD value that we are removing. They are 2 separate instances. 
  Yes, I am doing remove and then insert/add only if the event type is EDIT_FIELD.
   Else I am doing only insert/add if the event type is ADD_FIELD.

kview/src/main/java/dev/ikm/komet/kview/events/pattern/ShowPatternFormInBumpOutEvent.java
swaroopsalvi: I have added it to send the data from pattern details fields to the fields edit pane so that user can edit the values.
This is similar to the other event classes. Are we changing this implementation design across all the events? Can you please suggest the alternative ?

PR #79, opened 2024-10-09
Feature/iia 934 add stamp to pattern create

dholubek:
This adds the STAMP menu
updated the title panes in the menu to default to blank except the status is set to active,
this does touch concept but I tested that and its works
it also updates the "banner" at the top of the pattern to grab the FQN, publicId, and real stamp time instead of an in memory time placeholder

image

image


kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/PatternViewModel.java
```diff
@@ -216,7 +221,26 @@ public boolean createPattern() {
...
+
+    public String getPatternTitle() {
+        return ((EntityFacade) getPropertyValue(PATTERN)).description();
+    }
+
+    public Long getPatternLastStampTime() {
+        EntityFacade patternFacade = getPropertyValue(PATTERN);
+        Entity pattern = EntityService.get().getEntityFast(patternFacade.nid());
+        IntIdSet idSet = pattern.stampNids();
+        OptionalLong latestStamp = idSet.intStream().mapToLong(i -> ((StampEntity)EntityService.get().getEntityFast(i)).time()).max();
```
carldea, line 238: Can we use `getViewProperties().stampCalculator()`?

dholubek: updated to use the StampCalculator along with the code that @rbradley813 suggested to get the STAMP.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/PatternViewModel.java
```diff
@@ -216,7 +219,36 @@ public boolean createPattern() {
...
+        stampViewModel.setValue(TIME, stamp.time());
+        stampViewModel.setValue(AUTHOR, stamp.author());
+        stampViewModel.setValue(MODULE, stamp.module());
+        stampViewModel.setValue(PATH, stamp.path());
+    }
 
-        return composer.commitSession(session);
+    public String getPatternIdentifierText() {
+        EntityFacade patternFacade = getPropertyValue(PATTERN);
+        return String.valueOf(patternFacade.toProxy().publicId().asUuidList().getLastOptional().get());
```
carldea, line 252: Can we look at how our recent change from the concept window? 
It has code to also display the SNOMED and LOINC codes if present.

carldea (approved):
Looks good. 
I do have an optional suggestion / question. But if you want to defer to another ticket let me know. Since I will approve, go ahead an merge either way you decide above.

PR #85, opened 2024-10-13
Feature/ikm 20241013 hybrid reasoner addition

emays:
Adding hybrid reasoner option to classification


carldea (approved):
Looks good to me. Do we need to test (for jpms issues) before a merge?

PR #88, opened 2024-10-15
Bugfix/iia 1000 create new concept add sufficient set getting error

swaroopsalvi:
Using public Id to create concept instead of nid.

        PublicId conceptPublicId = PublicIds.newRandom();
        ConceptRecord conceptRecord = ConceptRecord.build(conceptPublicId.asUuidList().get(0), stampEntity.lastVersion());

        ConceptFacade conceptFacade = EntityProxy.Concept.make(conceptRecord.publicId()) ;


kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/ConceptViewModel.java
```diff
@@ -123,15 +120,15 @@ public boolean createConcept(EditCoordinateRecord editCoordinate) {
         save(); // View Model xfer values. does not save to the database but validates data and then copies data from properties to model values.
 
         // Validation errors will not create record.
-        if (getValidationMessages().size() > 0) {
+        if (!getValidationMessages().isEmpty()) {
```
carldea, line 123: Can we use the `hasErrorMsgs()` method?

carldea (approved):
Looks good. I had a comment, but we can fix it later. We were able to test and we need this fix right away. Thanks @swaroopsalvi !

PR #90, opened 2024-10-16
Added Permissions to release.yaml

OmOmofonmwan:


carldea:
@OmOmofonmwan Uyi, I'm not sure what to do this PR. Are we running into an issue with write permissions?

PR #92, opened 2024-10-18
Bumping up tinkar-core version 1.75.0 and cognitive 1.4.0

carldea:
Bumping up tinkar-core version 1.75.0 and cognitive 1.4.0


dholubek (approved):
looks good

PR #93, opened 2024-10-21
Feature / JPRO-33 Merge All Changes From komet-jpro to komet

besidev:
The purpose of this pull request is to merge all changes from the [komet-jpro](https://github.com/Sandec/komet-jpro) into main [komet](https://github.com/ikmdev/komet) repository, ensuring that Komet includes all the latest JPro-related updates to enable running it on the web.


besidev:
Hi @carldea Carl,

**Important Note:**
The `WebApp` class within the application module replicates the functionality of the App class while incorporating feature flags to enable running Komet on the web using JPro. Additionally, `WebApp` can be utilized to run the Komet application on desktop platforms (just set as `mainClass` property inside `javafx-maven-plugin` configuration the following value `dev.ikm.komet.application/dev.ikm.komet.app.WebApp`).

Currently, the `WebApp` class is maintained separately to facilitate the review and testing of JPro integration with the main Komet codebase. This separation ensures that the existing application entry point and structure remain unaffected.

Looking ahead, the plan is to streamline the application's architecture and unify the codebase by renaming the WebApp class to App, effectively replacing the existing App class.

cc/ @indritbeqiri

awesomejt-work:
@besidev ,

I've been tasked with reviewing the idea of containerization support within komet and what that might require from a CICD pipeline perspective. Reviewing this PR and the associated branch - I noticed the use of "jpro-maven-plugin" - but I cannot find this on Maven Central. Will this be made available or is this closed-source? We are trying to keep all dependencies open source.

Thanks,

Jason Taylor - Automation Team

jlgrock:
Yes please, if we can get that in a maven profile, and that's the only license concern we have (like the jpro dependencies) won't affect our overall Apache 2 distribution, then we should be good to merge in this PR with that profile .  We'll want to do some checks on our end.  

For anything related to the profile, please update the README to reflect changes as well.  But we are good with you spending the time getting the Sandec stuff pushed to maven central as well.  Please let me know if you need help with that (or check out our GitHub build yaml - we've worked through the automation of push-button releasing on the IKMDEV projects).

jlgrock:
also, please detail how are we going to run the `jpro-maven-plugin` when needed?

carldea:
@besidev @indritbeqiri 

> The WebApp class within the application module replicates the functionality...

Very good to know. Nice to be able to provide config info to run in the JPro Web context.

> Currently, the WebApp class is maintained separately...

No problem as far as I can tell. As a desktop app build I guess (maybe overkill) we could exclude using maven.

> Looking ahead, the plan is to streamline the application's architecture...

I think that's a good approach (yes/maybe?). Future effort we still may need to revisit our application launch life cycle. In past discussions, we spoke of changing the flow and wrt to the primary stage.

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:24.04
```
carldea, line 1: Could you try the following:
https://github.com/zulu-openjdk/zulu-openjdk/blob/master/ubuntu/21-latest/Dockerfile

kview/src/test/java/dev/ikm/komet/kview/mvvm/login/LoginTest.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    @BeforeEach
+    void setup() throws Exception {
+        // Register the primary stage in TestFX
+        FxToolkit.registerPrimaryStage();
+
+        // Create the users.ini file in the temporary directory
+        createUsersFile();
```
carldea, line 110: Question: if the file exists does it ignore the file?
Reasoning: Assumes the password can be changed manually and the file does not get overwritten.

docker/Dockerfile
besidev: Thanks for the suggestion Carl, we have updated the Dockerfile in this [commit](https://github.com/ikmdev/komet/pull/93/commits/7b80c7cc6428b7b7df971f2dce339663d36cc8dd).

carldea (approved):
Looks good to me!

kview/src/test/java/dev/ikm/komet/kview/mvvm/login/LoginTest.java
indritbeqiri: Correct! Soon we will make a commit related to the `LoginTest` in the case the file exists.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/login/LoginPageController.java
```diff
@@ -0,0 +1,132 @@
...
+            loginViewModel.authenticate(username, password);
+
+            // Clear the error labels
+            usernameErrorLabel.setText("");
+            passwordErrorLabel.setText("");
+
+            // Clear view model's values to remove passwords from memory
+            loginViewModel.setValue(USERNAME, "");
+            loginViewModel.setValue(PASSWORD, "");
+            loginViewModel.save(true);
```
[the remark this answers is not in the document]
besidev: Thanks for the detailed explanation, Carl! Using the `reset()` method is a better solution. However, as you mentioned, it also clears the values from the screen, which isn’t ideal when, for example, a wrong password is entered, we don’t want to clear the text fields in that case, forcing the user to re-enter their credentials. The good part is that TestFX unit test caught this behaviour change, which is great :).
To address [this](https://github.com/ikmdev/komet/pull/93/commits/f6e877409b892b93851c9796a72844e1fc6e26e9), I moved the same block into the Login view model so that the text fields are only cleared after successful authentication.

besidev:
Hi @jlgrock  Justin,

> Yes please, if we can get that in a maven profile, and that's the only license concern we have (like the jpro dependencies) won't affect our overall Apache 2 distribution, then we should be good to merge in this PR with that profile . We'll want to do some checks on our end...

We've [committed](https://github.com/ikmdev/komet/pull/93/commits/5288b35e9a3eef9e2cbc5cae865b906937fcaae7) the necessary changes by configuring `jpro-maven-plugin` as a Maven profile. Additionally, we've updated the JPro-related section of the README to reflect these changes.

Regarding pushing the Sandec components to Maven Central, we are currently working on that and will reach out if we need any assistance. Thank you in advance for your support.

Please let me know if there's anything else you need from our end!

PR #96, opened 2024-10-23
Fixes IIA-853: DnD drop zone visibility issue while dragging

dukke:


carldea (approved):
Excellent work!
This was a brain teaser for us. Great to see this elegant and simple fix. Thank you for explaining the mouse entered behavior of hovering over a child node and it's parent.

I forgot all about the `getPickResult()` method (been there since Java 8).

dukke:
Thanks Carl! Appreciate it! 🙂

The JavaFX API is so big that it’s impossible for someone to know all the
methods that exist. I just happen to probably have used getPickResult more
often…

PR #97, opened 2024-10-24
Bugfix / JPRO-67 Update JPro Server Stop Command for DEV Mode Following JPro Maven Plugin Profile Configuration Change

besidev:
**Background:** Komet utilizes the JPro Server for running JavaFX applications in a web environment, managed through the JPro Maven Plugin. Recently, the JPro Maven Plugin configuration was updated to use profile configurations.

**Issue:** After changing the JPro Maven Plugin to use profile configurations, the command responsible for stopping the JPro Server in DEV mode is no longer functioning as expected.

**Proposed Solution:** Update the JPro Server stop command configuration to align with the new profile-based setup of the JPro Maven Plugin.

### DIFF
--- a/application/src/main/java/dev/ikm/komet/app/WebApp.java
+++ b/application/src/main/java/dev/ikm/komet/app/WebApp.java
@@ -1146,11 +1146,11 @@ public class WebApp extends Application {
             final String[] stopScriptArgs;
             final CommandRunner stopCommandRunner;
             final File workingDir;
-            if ("dev".equals(jproMode)) {
+            if ("dev".equalsIgnoreCase(jproMode)) {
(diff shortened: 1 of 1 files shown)


carldea (approved):
Looks great! On our side we need our testers to know how to manually test the stop.

besidev:
Thank you, Carl!

We’ll provide your testers with instructions on how to manually test the `start` and `stop` commands. JPro supports these commands both in development mode using the JPro Maven Plugin and in production deployments via the release zip, located inside the `bin` directory.

FYI: While these commands are primarily used in production, during development the `jpro:run` command is mostly utilized. It runs the server in the foreground, which is very useful for reading logs and monitoring terminal output. This effectively replicates the functionality of the `javafx:run` command, even though they are not the same.

PR #98, opened 2024-10-24
Feature / JPRO-68 Run Login Tests When Users File Exists

indritbeqiri:
The login tests are designed to verify the application's user authentication mechanisms. Currently, these tests rely on a users file containing predefined user credentials. This setup is typically used in Continuous Integration (CI) environments or when the users file does not exist.

This update introduces support for scenarios where the users file already exists and may have been modified manually. In such cases, the login tests will retrieve user credentials from the existing users file instead of using the predefined credentials. This enhancement ensures that:

**Data Integrity:** Validates that the existing users file contains correct and uncorrupted user data.

**Validation:** Confirms that the user data within the users file is accurate and usable for testing purposes.


carldea (approved):
It looks good to me.
Thank you.

indritbeqiri:
Thank you Carl!

PR #104, opened 2024-10-30
Bugfix/iia 1028 unable to add multiple fields from add field button on properties panel

swaroopsalvi:
Added field order for properties ADD field event.


dholubek (approved):
looks good

PR #105, opened 2024-10-31
Bugfix/iia 470 the case significance and status drop own fields shows duplicate

swaroopsalvi:


dholubek (approved):
looks good

carldea (approved):
Looks good, I will merge.

PR #110, opened 2024-11-06
Added support for rendering role-chains (property sequence implication).

kec:


carldea (approved):
Looks reasonably safe, however we'll need to regression test it afterwards.

PR #111, opened 2024-11-06
Bugfix / JPRO-72 Unable to load `kview.css` when deploying Komet Application via JPro to Production (testing) Server

besidev:
When deploying the Komet application to the production (testing) server using the JPro release zip, the `kview.css` file fails to load. This issue arises because the module descriptor of the kview module, where `kview.css` is located, is missing an `opens` directive.

### DIFF
--- a/application/src/main/java/dev/ikm/komet/app/util/CssFile.java
+++ b/application/src/main/java/dev/ikm/komet/app/util/CssFile.java
@@ -1,0 +1,119 @@
+package dev.ikm.komet.app.util;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Enum representing supported CSS files along with their associated metadata.
+ * <p>
+ * Each enum constant encapsulates the CSS file name, module name, resource path within the module
+ * and the base directory for resources.
+ * </p>
+ * <p>
+ * To add a new CSS file, declare a new enum constant with the appropriate metadata.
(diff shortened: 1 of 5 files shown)


application/src/main/java/dev/ikm/komet/app/util/CssFile.java
```diff
@@ -0,0 +1,119 @@
...
+            "komet.css",
+            "dev.ikm.komet.framework",
+            "dev/ikm/komet/framework/graphics/komet.css",
+            Paths.get("framework", "src", "main", "resources")
+    ),
+
+    KVIEW_CSS(
+            "kview.css",
+            "dev.ikm.komet.kview",
+            "dev/ikm/komet/kview/mvvm/view/kview.css",
```
carldea, line 31: More of a question: since the file is on line 29, should line 31 contain the file name?

carldea (approved):
Looks good to me!
I can merge, but only had a small question?

carldea:
@besidev Great work btw, for the sake of time, I will merge it since it was tested. My question was very minor.

application/src/main/java/dev/ikm/komet/app/util/CssFile.java
besidev: Correct, we can re-use the file name from line 29 to line 31. Making the required change.

PR #115, opened 2024-11-08
Bugfix / JPRO-73 (Mac Only) JPRO-70 Merge Breaks Production Release

besidev:


carldea (approved):
@besidev 
Thank you! I tested the install in the PR's build and was able to install and run successfully as an installed app for MacOS arm64.

carldea:
Thank you also for the quick turnaround.

PR #117, opened 2024-11-08
Feature / JPRO-66 Push JPro Dependencies to Maven Central

indritbeqiri:
Ensure that all necessary JPro-related libraries required for the Komet project are available on Maven Central.

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -90,7 +90,7 @@
         </dependency>
         <!-- Start JPro related dependencies -->
         <dependency>
-            <groupId>com.sandec.jpro</groupId>
+            <groupId>one.jpro</groupId>
             <artifactId>jpro-webapi</artifactId>
             <scope>compile</scope>
         </dependency>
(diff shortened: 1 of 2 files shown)


carldea (approved):
Looks good to me!

indritbeqiri:
Thank you, Carl. Please conduct all necessary tests prior to merging.

application/pom.xml
```diff
@@ -481,6 +481,18 @@
             <activation>
                 <activeByDefault>false</activeByDefault>
             </activation>
+            <pluginRepositories>
```
jlgrock, line 484: Is this artifactory necessary?  I believe we have pushed all of the needed artifacts to maven central, if I recall correctly.

FlorianKirmaier: We pushed the WebAPI to Mavencentral.
However, the repository is needed for JPro itself.
We are still talking internally about whether we should move the JPro plugins.
But that shouldn't matter too much.
The repositories are hidden under a Profile.
When JPro is used, our repository is used anyway.
It shouldn't influence the desktop build.

I Hope that's fine on your side.

carldea: @FlorianKirmaier 
I believe this is isolated in a profile and likely fine. Our main concern is making sure it does not prevent customers from building Komet publicly.

We will merge this PR for now 👍🏼 . And later revisit with a follow on ticket if or when your team decides to move libraries.

PR #118, opened 2024-11-08
Bugfix/IIA-1055: scroll not working pattern navigation

dukke:


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
```diff
@@ -1216,7 +1216,7 @@ private void loadNavigationPanel(PublicIdStringKey<ActivityStream> navigationAct
                                 patternNavViewModel.setPropertyValue(VIEW_PROPERTIES, viewProperties)
                         );
         JFXNode<Pane, ConceptPatternNavController> conPatJFXNode = FXMLMvvmLoader.make(patternConceptConfig);
-        patternConceptNavigationPanel = conPatJFXNode.node();
+        patternConceptNavigationPanel = (StackPane) conPatJFXNode.node();
```
carldea, line 1219: Can we remove the down cast of a `StackPane`?
Can we declare `conPatJFXNode` like the following?
```java
JFXNode<StackPane, ConceptPatternNavController> conPatJFXNode = ...
```
This helps avoid a downcast.

dukke: Good point that was a stupid miss from me. I've changed it. Thanks

carldea (approved):
Looks Good!

PR #119, opened 2024-11-08
Feature/iia 769 implement functionality to modify fully qualified name on pattern during creation

swaroopsalvi:


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PropertiesController.java
```diff
@@ -317,10 +319,17 @@ private void setupDescriptionNamePane(EvtType eventType) {
             });
         } else if (eventType == SHOW_EDIT_FQN) {
             descrConfig.updateViewModel("descrNameViewModel", (descrNameViewModel) -> {
```
carldea, line 321: Can we use the new features of Cognitive(mvvm library) to update a view model?
In line 210 Cognitive(mvvm) now allows you to obtain the view model from the JFXNode **after** it was created (via make()). 

- Functional style: https://github.com/carldea/cognitive/wiki#jfxnodes-updateviewmodel
- Imperative style: https://github.com/carldea/cognitive/wiki#jfxnodes-getviewmodel

carldea (approved):
Looks good to me. Thank you for using the updated API with respect to obtaining view models from JFXNode objects.

PR #122, opened 2024-11-12
Bumped up versions for cognitive v1.5.1 and axonic v1.1.5 

carldea:
Bumped up versions for cognitive v1.5.1 and axonic v1.1.5 slf4j logging services conflict with already loaded loggers.


rbradley813 (approved):
Looks good!

PR #123, opened 2024-11-12
bugfix/IIA-1063: Concept bumpout is not filling the whole height

dukke:
Fixes this issue: https://ikmdev.atlassian.net/browse/IIA-1063


carldea (approved):
@dukke  Excellent work!
Interesting how you bound the view port of the scroll pane.

PR #126, opened 2024-11-12
bugfix/IIA-1069: Top part of Concept bump out doesn't look as the Figma mockup

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1069


carldea (approved):
Looks good to me!

PR #130, opened 2024-11-14
feature/IIA-781: Create a mock-up of the pattern instance tab

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-781

Here's the "Instances panel" (note the ticket is only to create the mockup, so population with real data will be done on another ticket):
image


dholubek (changes requested):
can we add a screenshot to the PR comments, since it is a mockup?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PropertiesController.java
```diff
@@ -210,6 +216,15 @@ private void initialize() {
         patternFieldsPane = patternFieldsJFXNode.node();
         patternFieldsController.setViewProperties(getViewProperties());
 
+        // +-----------------------------------
+        // ! Instances Toggle selected
+        // +-----------------------------------
+        Config instancesConfig = new Config(INSTANCES_URL)
+                .addNamedViewModel(new NamedVm("patternPropertiesViewModel", patternPropertiesViewModel));
```
carldea, line 223: This isn't recommended as the parent controller really owns the `patternPropertiesViewModel`. The `patternPropertiesViewModel` variable name does not exist inside the `InstanceController`. I'm surprised it didn't throw an runtime exception. 
A quick fix would be:
remove the line: 223. or call 👍🏼 
```java
  .updateViewModel("simpleViewModel",  (viewModel) -> { 
      // viewModel.addProperty(property, value);
  } );
```

When injecting view models the following can are allowed:
- A variable name must exist in controller code via `@InjectViewModel`. a NamedVm is a key value of the variable name and view model respectively.
- Caller can construct their own instance to be injected, but the variable name must exist. FXMLMvvm.make() will  use caller's and not instanciate a new one.

dukke: This is not the InstancesController, it's the PropertiesController class.

dukke:
> can we add a screenshot to the PR comments, since it is a mockup?

good point. Done.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PropertiesController.java
carldea: Maybe i'm missing what you are saying but below shows line 222 inside of `PropertiesController`:
```java
Config instancesConfig = new Config(INSTANCES_URL)
                .addNamedViewModel(new NamedVm("patternPropertiesViewModel", patternPropertiesViewModel));
```
Is the above code stating `new Config(INSTANCES_URL)`? That is the fxml and controller being created via `FXMLMvvm.make()`. Here it is the code to create the FXML and InstancesController i thought (child window of PropertiesController).

carldea (approved):
Thank you. Glad we went over the usage of Cognitive (MVVM).

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/InstancesController.java
```diff
@@ -0,0 +1,15 @@
+package dev.ikm.komet.kview.mvvm.view.pattern;
+
+import javafx.fxml.FXML;
+import javafx.scene.control.ListView;
+import org.carlfx.cognitive.loader.InjectViewModel;
+import org.carlfx.cognitive.viewmodel.SimpleViewModel;
+
+public class InstancesController {
+    @InjectViewModel
+    private SimpleViewModel simpleViewModel;
```
dholubek, line 12: you could change this to be the patternViewModel 

`@InjectViewModel
private PatternViewModel patternViewModel`

and match it with what is in the config down below

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PropertiesController.java
dholubek: And this would match what is inside the InstancesController

`Config instancesConfig = new Config(INSTANCES_URL)
                .addNamedViewModel(new NamedVm("patternViewModel", patternViewModel));`

dholubek (approved):
I'm restating what Carl is saying, but we don't necessarily need the viewModels set up in this story, it can be done in a future story.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/InstancesController.java
carldea: Yes we could do this in the other ticket. Nice suggestion. We could share the parent's view model. My only issue, is if the observable list of instances is very large (1M+) it might not be a good idea to store it. (this is basically read-only rows of data) i think :-).

PR #133, opened 2024-11-14
Bugfix/iia 1086 the pattern navigator is not getting refreshed after creating a new pattern and save

swaroopsalvi:


kview/src/main/java/dev/ikm/komet/kview/events/EventTopics.java
```diff
@@ -18,4 +18,5 @@
 public enum EventTopics {
     JOURNAL_TOPIC,
     USER_TOPIC,
+    SAVE_PATTERN_TOPIC,
```
dholubek, line 21: (optional) we didn't use the JOURNAL_TOPIC?  I don't know if LIDR or Concept would need to listen to it but they could listen on the JOURNAL_TOPIC...?

dholubek (commented):
had one comment... maybe we should ask @carldea

dholubek (approved):
approved

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -766,6 +769,8 @@ private void savePattern(ActionEvent actionEvent) {
         LOG.info(isValidSave ? "success" : "failed");
         updatePatternBanner();
         updateView();
+        actionEvent.consume();
```
dukke, line 772: I don't think we should consume events (mouse events, etc) as a standard practice. 
The problem is that if we want to listen to this type of event somewhere up on the hierarchy by adding an event handler you won't be able to because the event was consumed. Also, there's no real advantage to do that in most cases.

dukke (commented):
Nicely done! Just had one minor comment to make.

PR #137, opened 2024-11-18
Feature/iia 994 summon pattern via dnd

dholubek:
https://ikmdev.atlassian.net/browse/IIA-994


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/navigation/ConceptPatternNavController.java
```diff
@@ -327,4 +373,39 @@ protected void updateItem(Object item, boolean empty) {
...
+        Objects.requireNonNull(node, "The node must not be null.");
+        Objects.requireNonNull(entity, "The entity must not be null.");
+
+        // Associate the node with the entity's public ID and type for later retrieval or identification
+        node.setUserData(new DragAndDropInfo(dropType, entity.publicId()));
+
+        // Set up the drag detection event handler
+        node.setOnDragDetected(mouseEvent -> {
+            // Initiate a drag-and-drop gesture with copy or move transfer mode
+            Dragboard dragboard = node.startDragAndDrop(TransferMode.COPY_OR_MOVE);
```
dukke, line 386: Do we want to start a platform drag and drop gesture here? This is best used for instance if you want to drag and drop between applications...
For drag and dropping within the same application we can call setFullDrag (mouse events with a setFullDrag are also different - MouseDragEvent class)

dukke (commented):
Had one comment to make, the rest looks good

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/navigation/ConceptPatternNavController.java
dholubek: where would we call setFullDrag?  @dukke could you provide a code example because I don't see where that method lives... this is currently what NextGenSearch has for drag and drop.

dholubek: @carldea can we leave as is and refactor later?

swaroopsalvi: I think it will be good to catch early and fix early, we might drop the ball to fix this later since the refactor list seems to be getting long.

swaroopsalvi (commented):
Its better to catch early and fix early if we can rather than  do it as part of refactor.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/navigation/ConceptPatternNavController.java
carldea: Let's create another refactoring ticket. Great explanation @dukke .

carldea (approved):
Looks good to me.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/navigation/ConceptPatternNavController.java
dholubek: https://ikmdev.atlassian.net/browse/IIA-1128 is the refactoring ticket

PR #138, opened 2024-11-19
Bugfix/IIA-1065: Scroll works differently on Concept bumpout depending on which toggle is selected

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1065

It's ready for review... video of how it's working now and how it was working before is available in the Jira ticket.


dholubek (approved):
looks good

PR #142, opened 2024-11-21
Feature/iia 782 pattern instances data

dholubek:
https://ikmdev.atlassian.net/browse/IIA-782

image

image


dukke:
Great you're grabbing the subtitle from the database now.

A couple general comments:
1 - perhaps we could attach a picture of when there are just a few items? Just to show reviewers how it looks when that happens (no scrollbar, empty ListView space)?
2 - There's sometimes an extra space at the beginning of the title of an instance, which makes the title and the subtitle get missaligned:
image
Is that because the string that come from the backend has an extra space at the beginning? Can we possibly remove that by calling `stripLeading()` on the string (removes extra leading space on a String) before populating the List Cell?

swaroopsalvi (approved):
Looks good.

dholubek:
> Great you're grabbing the subtitle from the database now.
> 
> All good just a couple general comments: 1 - perhaps we could attach a picture of when there are just a few items? Just to show reviewers how it looks when that happens (no scrollbar, empty ListView space)? 2 - There's sometimes an extra space at the beginning of the title of an instance, which makes the title and the subtitle get missaligned: image Is that because the string that come from the backend has an extra space at the beginning? Can we possibly remove that by calling `stripLeading()` on the string (removes extra leading space on a String) before populating the List Cell?

it is the data that has the leading space

PR #143, opened 2024-11-21
Feature/iia 927 and iia 1111 implement functionality to edit pattern definition and stamp exception bugfix

swaroopsalvi:
Code to fix STAMP saving bug, and Edit pattern definition

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDefinitionController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDefinitionController.java
@@ -15,21 +15,12 @@
  */
 package dev.ikm.komet.kview.mvvm.view.pattern;
 
-import static dev.ikm.komet.kview.events.pattern.PatternDefinitionEvent.PATTERN_DEFINITION;
-import static dev.ikm.komet.kview.events.pattern.PropertyPanelEvent.CLOSE_PANEL;
-import static dev.ikm.komet.kview.events.pattern.PropertyPanelEvent.DEFINITION_CONFIRMATION;
-import static dev.ikm.komet.kview.mvvm.viewmodel.DescrNameViewModel.IS_INVALID;
-import static dev.ikm.komet.kview.mvvm.viewmodel.PatternDefinitionViewModel.MEANING_ENTITY;
-import static dev.ikm.komet.kview.mvvm.viewmodel.PatternDefinitionViewModel.PURPOSE_ENTITY;
-import static dev.ikm.komet.kview.mvvm.viewmodel.PatternViewModel.PATTERN_TOPIC;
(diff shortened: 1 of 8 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDefinitionController.java
```diff
@@ -155,30 +148,34 @@ private void initialize() {
...
+        ObjectProperty<ConceptEntity> purposeProp = patternDefinitionViewModel.getProperty(PURPOSE_ENTITY);
+        ObjectProperty<ConceptEntity> meaningProp = patternDefinitionViewModel.getProperty(MEANING_ENTITY);
         purposeProp.addListener(fieldsValidationListener);
+        purposeProp.addListener((obs, oldVal, newVal) -> {
+            if(newVal !=null){
+                addPurposeToForm(newVal);
+            }else{
+                removePurposeForm();
+            }
+        });
```
dukke, line 170: if we're not using the old value perhaps an InvalidationListener would do? (slightly better performance)

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDefinitionController.java
```diff
@@ -155,30 +148,34 @@ private void initialize() {
...
         meaningProp.addListener(fieldsValidationListener);
+        meaningProp.addListener((obs, oldVal, newVal) -> {
+            if(newVal !=null){
+                addMeaningToForm(newVal);
+            }else{
+                removeMeaningForm();
+            }
+        });
```
dukke, line 178: if we're not using the old value perhaps an InvalidationListener would do? (slightly better performance)

dukke (commented):
Just had 1 comment (2 comments but same issue). Other than just those 2 all looks good to me!

dholubek (commented):
just one question, it might be better that it replaces on drop.  I just want to make sure it still works okay without an error or mashing two in there when it isn't supposed to.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDefinitionController.java
```diff
@@ -155,30 +148,34 @@ private void initialize() {
         doneButton.disableProperty().bind(patternDefinitionViewModel.getProperty(IS_INVALID));
 
         setupDragNDrop(purposeStackPane, (publicId) -> {
-            // check to see if a pattern > purpose was already dragged into the purpose section before saving
-            // to the view model
-            if (patternDefinitionViewModel.getPropertyValue(PURPOSE_ENTITY) == null) {
```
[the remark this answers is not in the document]
dholubek: this works... I just tried it.  going to approving

dholubek (approved):
looks good

PR #146, opened 2024-11-22
Feature/iia 927 implement functionality to edit pattern definition and other broken defects

swaroopsalvi:
Fixed drag and drop user data issue and create new pattern Null Pointer exception bug that got introduced due to InstanceController.


carldea (approved):
Looks good to me.


I do caution the use of embedding data in UI elements. But this will be fine for now. Selection or clicking on items can be managed in view models.

PR #149, opened 2024-11-26
Bugfix/iia 1134 globe icon button is throwing an exception while editing pattern and saving the updates

swaroopsalvi:


dholubek (approved):
looks good

PR #150, opened 2024-11-28
BugFix/IIA-1087: The drop down values in komet app are not sorted by alphabetical order

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1087

1. I've fixed this on some screens however I don't know all the places where this needs to be fixed. I left a comment in the Jira ticket asking that but haven't received a reply yet
2. I've created a new "custom control" called SortedComboBox and I've put it in the same package as the package Gluon is using on their custom controls issue. Whenever we need values in a ComboBox to be sorted we'd use this SortedComboBox

image


kview/src/main/java/dev/ikm/komet/kview/controls/SortedComboBox.java
```diff
@@ -0,0 +1,39 @@
+package dev.ikm.komet.kview.controls;
+
+import dev.ikm.tinkar.common.util.text.NaturalOrder;
+import javafx.beans.Observable;
+import javafx.collections.ListChangeListener;
+import javafx.collections.ObservableList;
+import javafx.scene.control.ComboBox;
+
+public class SortedComboBox<T> extends ComboBox<T> {
```
carldea, line 9: Can we document the class?
Is the sorting ascending only?

dukke: Added javdoc to the SortedComboBox.

We were instructed in the ticket that the client wanted us to use the function `NaturalOrder.getObjectComparator()`. 
I suppose it's ascending but I guess that could change if the client decides to change the implementation of that method, so it's not totally clear to me what the client wants exactly except that we need to call that method and have the combobox contents display in some order.

swaroopsalvi (commented):
We should be doing the sorting for patterns combo boxes as well right ?
and update PropertiesController.java in the   *.pattern.PropertiesController.

dukke:
> Thank you! Nice work. I like the consistency with having sorted options.

Thanks, glad you like it Carl! Yes, I think it makes it easier for a user to go through a long list and find the one item he is looking for...

swaroopsalvi (commented):
@dukke 
   The following files can be looked for this:
      **_dev.ikm.komet.kview.mvvm.view.pattern.PatternController.java_**
       In this file there are references to the fxml and their respective controllers.

     `  private static final URL PATTERN_DESCRIPTION_FXML_URL = DescriptionNameController.class.getResource("description-name.fxml");

    private static final URL PATTERN_FIELDS_FXML_URL = PatternFieldsController.class.getResource("pattern-fields.fxml");
`
The other place where this sorting needs to be in the same order as that of the StampEditController and edit-stamp.fxml.

That said, I think another better approca will be to sort the data in the helper class **

_dev.ikm.komet.kview.mvvm.model.DataModelHelper.java_

**

   Here we are calling the:
   public static Set<ConceptEntity> fetchDescendentsOfConcept(ViewProperties viewProperties, PublicId publicId) { ...}
    and we can implment the sort method here itself. this will reduce the load of the UI to do sorting of data.

So the method will having a small change:  the sorted() method of Stream.

      public static Set<ConceptEntity> fetchDescendentsOfConcept(ViewProperties viewProperties, PublicId publicId) {
        IntIdSet decendents = viewProperties.calculator().descendentsOf(EntityService.get().nidForPublicId(publicId));
        Set<ConceptEntity> allDecendents = decendents.intStream().sorted()
                .mapToObj(decendentNid -> (ConceptEntity) Entity.getFast(decendentNid))
                .collect(Collectors.toSet());
        return allDecendents;
    }

dukke:
> We should be doing the sorting for patterns combo boxes as well right ? and update PropertiesController.java in the *.pattern.PropertiesController.

Added sorting to the places @swaroopsalvi suggested: Pattern Window ComboBoxes and Stamp radiobutton collections

kview/src/main/java/dev/ikm/komet/kview/controls/SortedComboBox.java
kec: Natural order can be used to sort ascending or descending by a simple sign change in the sort result

dukke: @kec Oh I see. Nice! Thanks for clarifying Dr. Campbell! 🙂

PR #154, opened 2024-12-03
BugFix/IIA-1136: Shrinking a window can make its contents get cut out

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1136

## Some Remarks
There's a bunch of code duplication between Lidr, Pattern and Concept Window. This particular bug happens on all 3 windows and because of the code duplication had to be fixed using the same fix (same code) on 3 different places.

@carldea @dholubek @swaroopsalvi 
Perhaps it would be interesting to create a ticket to refactor these classes and remove the code duplication?

### Actual Behavior

When shrinking a Window its contents can get cut out.

For example, when shrinking the Concept Window the buttons on the bottom of “Add Description: Fully Qualified Name” disappear (video bellow shows this issue happening). 

https://github.com/user-attachments/assets/f3881c41-f6b9-4ff5-b91c-53ea5996a9e0

This also happens for example in the “Add other name” pane and on Lidr Window when “Add/Edit” tab is selected.

This bug can happen on any type of Window (Concept or Lidr or Pattern) depending on the contents of the pane on its right. 

### Expected Behavior

Content of Window shouldn’t get cut out.

### Steps to Reproduce

1 - Click the plus bottom at the top and click to create new Concept Window
2- Click the Toggle Switch button so that “Add Description: Fully Qualified Name” is showing
3- Resize Window and shrink it vertically 
4 - There will be a point where the button at the bottom disappear.


swaroopsalvi (approved):
Approve

PR #156, opened 2024-12-05
Feature / JPRO-25 Build TestFX POC for Automated Testing

indritbeqiri:
https://ikmdev.atlassian.net/browse/JPRO-25

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-P!codeQuality
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -628,9 +628,6 @@
                             <includes>
                                 <include>**/*IT.java</include>
                             </includes>
-                            <systemPropertyVariables>
-                                <target.data.directory>${project.build.directory}/data</target.data.directory>
-                            </systemPropertyVariables>
                             <argLine>
                                 -Dtestfx.headless=true
                                 -Djava.awt.headless=true
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -167,33 +167,6 @@
                         </options>
                     </configuration>
                 </plugin>
-<!--                <plugin>-->
-<!--                    <groupId>org.apache.maven.plugins</groupId>-->
-<!--                    <artifactId>maven-dependency-plugin</artifactId>-->
-<!--                    <executions>-->
-<!--                        <execution>-->
-<!--                            <id>copy-data</id>-->
-<!--                            <phase>pre-integration-test</phase>-->
-<!--                            <goals>-->
-<!--                                <goal>copy</goal>-->
-<!--                            </goals>-->
-<!--                        </execution>-->
-<!--                    </executions>-->
-<!--                    <configuration>-->
-<!--                        <artifactItems>-->
-<!--                            <artifactItem>-->
-<!--                                <groupId>dev.ikm.tinkar.data</groupId>-->
-<!--                                <artifactId>tinkar-starter-data-reasoned</artifactId>-->
-<!--                                <classifier>pb</classifier>-->
-<!--                                <type>zip</type>-->
-<!--                                <overWrite>false</overWrite>-->
-<!--                                <outputDirectory>${user.home}/Solor</outputDirectory>-->
-<!--                            </artifactItem>-->
-<!--                        </artifactItems>-->
-<!--                        <overWriteReleases>false</overWriteReleases>-->
-<!--                        <overWriteSnapshots>true</overWriteSnapshots>-->
-<!--                    </configuration>-->
-<!--                </plugin>-->
             </plugins>
         </pluginManagement>
         <plugins>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -102,6 +102,20 @@
             <groupId>one.jpro.platform</groupId>
             <artifactId>jpro-internal-util</artifactId>
         </dependency>
+        <!--        testing -->
+        <dependency>
+            <groupId>one.jpro.platform.jpms</groupId>
+            <artifactId>testfx-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>one.jpro.platform.jpms</groupId>
+            <artifactId>testfx-junit5</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>one.jpro.platform.jpms</groupId>
+            <artifactId>openjfx-monocle</artifactId>
+            <scope>test</scope>
+        </dependency>
         <!-- End JPro related dependencies -->
         <dependency>
             <groupId>${tinkar-core.groupId}</groupId>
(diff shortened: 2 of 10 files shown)


carldea (dismissed):
@indritbeqiri 
This looks great. I'm very impressed!

I will approve but will need to test the branch before the merge. We want to make sure things are good on all environments.

We will need to have further discussions and testing in the various environments and prepare the QA/Dev team on the appropriate way to skip and how to convert test scripts into TestFX tests.

carldea:
@indritbeqiri  Indrit, I backed out the approval let me explain.

**Here are some of our concerns:**
- We need to ensure code quality (SonarQube) is turned on (profile or flag).
- We want to understand artifacts are being used in a headless mode. I noticed the maven coordinate having GAV group id from JPro. What is the state of Monocle should these changes go back to the owners or is it okay with this forked release?
- We should work with our Automation team as it relates to Runtime, Module, Build, Code quality issue.
- Verify the tests are using the **ephemeral** datastore as opposed to a filesystem type database.
- We'll have to discuss with the automation team about when these types of tests should be run.
- What are next steps once the PR gets merged. Possibly Developer workflow training (on our side).

.mvn/maven.config
```diff
@@ -0,0 +1 @@
+-Pits
```
jlgrock, line 1: OK, let's remove this entirely.  If it needs to be applied every time, you would just make the profile active by default.

application/pom.xml
```diff
@@ -530,5 +544,76 @@
...
+                        </executions>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>dev.ikm.tinkar.data</groupId>
+                                    <artifactId>tinkar-starter-data-reasoned</artifactId>
+                                    <classifier>pb</classifier>
+                                    <type>zip</type>
+                                    <overWrite>false</overWrite>
+                                    <outputDirectory>${user.home}/Solor</outputDirectory>
```
jlgrock, line 574: This should not be in the home directory, but should be in the maven target directory.  This allows for cleanup of the artifacts, whereas somewhere in the home dir does not

pom.xml
```diff
@@ -602,6 +621,11 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven-surefire-plugin.version}</version>
                 </plugin>
+                <plugin>
```
jlgrock, line 624: both the maven surefire and failsafe plugins need to be inherited from the parent pom and should not need to be listed here.

pom.xml
```diff
@@ -89,9 +89,11 @@
         <jpro-auth-core.version>${jpro-platform.version}-ikmdev</jpro-auth-core.version>
 
         <!-- TestFX Variables -->
+        <tinkar-starter-data-reasoned.version>1.0.0</tinkar-starter-data-reasoned.version>
         <testfx.version>4.0.18-jpms</testfx.version>
         <openjfx-monocle.version>21.0.2-jpms</openjfx-monocle.version>
         <maven-surefire-plugin.version>3.5.0</maven-surefire-plugin.version>
```
jlgrock, line 95: we need to get the parent pom to be updated to use this version of surefire/failsafe rather than changing this for individual projects.  We want to stay consistent on build tooling

pom.xml
indritbeqiri: Correct! I was just finishing the last tests before making the commits. I'm going to use the `codeQuality` profile to call the maven surefire and failsafe plugins.

application/pom.xml
```diff
@@ -530,5 +544,76 @@
...
+                            <includes>
+                                <include>**/*IT.java</include>
+                            </includes>
+                            <argLine>
+                                -Dtestfx.headless=true
+                                -Djava.awt.headless=true
+                                ${commonTestFXArgs}
+                            </argLine>
+                            <useModulePath>true</useModulePath>
+                            <forkCount>1</forkCount>
```
jlgrock, line 610: I believe we want parallelism in standard integration tests and we can accept no parallelism in monocle tests.  For that, I believe we need to push this down to execution level configuration.

jlgrock:
hey, added some comments to fix.  I believe this should be minor in the grand scheme of things.  Some of it may be carried over from our code, but while we're in there, we're going to have you work on it.  Appreciate it!

application/pom.xml
indritbeqiri: That was the initial implementation but somehow it failed during the CI builds. I'll return back to the initial implementation using the target directory and try it again, but after have organized have reconfigured the maven surefire and failsafe plugins.

indritbeqiri:
> hey, added some comments to fix. I believe this should be minor in the grand scheme of things. Some of it may be carried over from our code, but while we're in there, we're going to have you work on it. Appreciate it!

Thank you Justin for the suggestions. I'm working on it and soon will commit the needed changes making sure that everything works correctly.

indritbeqiri:
> @indritbeqiri Indrit, I backed out the approval let me explain.
> 
> **Here are some of our concerns:**
> 
> * We need to ensure code quality (SonarQube) is turned on (profile or flag).
> * We want to understand artifacts are being used in a headless mode. I noticed the maven coordinate having GAV group id from JPro. What is the state of Monocle should these changes go back to the owners or is it okay with this forked release?
> * We should work with our Automation team as it relates to Runtime, Module, Build, Code quality issue.
> * Verify the tests are using the **ephemeral** datastore as opposed to a filesystem type database.
> * We'll have to discuss with the automation team about when these types of tests should be run.
> * What are next steps once the PR gets merged. Possibly Developer workflow training (on our side).

@carldea Carl,

Thank you for the detailed feedback. I fully agree with your points and understand the importance of addressing them.

Artifacts for the TestFX framework (including headless mode):

- org.testfx:testfx-core:4.0.18
- org.testfx:testfx-junit5:4.0.18
- org.testfx:openjfx-monocle:21.0.2

These artifacts are currently under the JPro GAV because we needed to modularize, integrate, and configure them within the ikmdev ecosystem, testing until we arrived at the right configuration. Having now demonstrated through this POC that it is possible to run automated integration tests both graphically and in headless mode, we can proceed to create PRs that add these libraries to the `ikmdev/jpms-deps` repository, where other non-JPMS-compliant libraries are maintained.

**Test Configuration Updates:**
After reviewing how tests currently run across the ikmdev ecosystem, I will commit changes to use Maven Surefire and Failsafe plugin configurations within the `codeQuality` profile in the` java-parent` `pom.xml`. This approach will help avoid conflicts and ensure more predictable behavior for these test plugins going forward.

Thank you and please let me know if you have any further suggestions.

.mvn/maven.config
indritbeqiri: The initial plan was to activate this profile only when needed. I reintroduced it here, not to make it active by default, but rather to override the existing Maven Failsafe Plugin configured in the `codeQuality` profile of the `java-parent` pom.xml, as the integration tests were running without utilizing the `its` profile’s configurations. Making it active by default did not resolve the issue. Therefore, I’m removing the Maven Failsafe Plugin from the `its` profile and relying solely on the one defined in the `codeQuality` profile within the `java-parent` pom.xml. I believe this is the correct approach to avoid conflicts and unpredictable behavior.

indritbeqiri:
@jlgrock @carldea 

Please refer to the [comment](https://ikmdev.atlassian.net/browse/JPRO-25?focusedCommentId=14462) on Jira ticket JPro-25 for the latest updates regarding this PR.

pom.xml
indritbeqiri: Correct! Now that the `java-parent` has been updated to the latest version, they are no longer needed here.

application/pom.xml
indritbeqiri: Instead, using a negated property for default activation worked. Moreover, this approach provides a simple, predictable way to skip the execution of the profile whenever necessary.

.mvn/maven.config
indritbeqiri: Instead, using a negated property for default activation worked. Moreover, this approach provides a simple, predictable way to skip the execution of the profile whenever necessary.

indritbeqiri:
The TestFX integration test failed with the following error:

TestFX integration test error


It appears that the `com.pixelduke:fxcomponents` library is using a different JavaFX version (`11.0.2`):
Dependecies used from the fxcomponenets library

This mismatch can lead to compatibility issues, causing methods like `treeShowingProperty` to become unavailable:
`java.lang.NoSuchMethodError: 'javafx.beans.binding.BooleanExpression com.sun.javafx.scene.NodeHelper.treeShowingProperty(javafx.scene.Node)'`

To resolve this issue, a temporary solution to pass the integration test is to exclude the JavaFX dependencies from `com.pixelduke:fxcomponents`. This ensures that the application exclusively uses the project's JavaFX version (`21.0.4`), preventing any version conflicts.

indritbeqiri:
@carldea @jlgrock 

In the latest commit, TestFX unit and integration tests have been changed to run on-demand instead of by default. Additionally, the README has been updated with instructions on how to execute these tests when needed.

Since the tests are no longer run automatically, a proposed suggestion is to merge this PR. It can be decided later when and where these tests should be executed. The important part is that TestFX is properly configured, especially for running integration tests headless in the CI using the module system, which was the main goal of this proof of concept (POC).

PR #161, opened 2024-12-11
Feature/iia 1106 enable users to modify patterns after creation

swaroopsalvi:


dukke:
Some of what's being used here from Komet I'm not yet familiar with but other than that, from my end all looks good! 👍

dukke (approved):

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/PatternViewModel.java
```diff
@@ -307,43 +311,47 @@ public boolean createPattern() {
...
+                    .text(otherName.getNameText())
+                    .caseSignificance(otherName.getCaseSignificance().toProxy());
+            if (isEdit) {
+                HashMap<DescrName, SemanticEntityVersion> regularNamesMap = getPropertyValue(REGULAR_NAMES);
+                if(regularNamesMap != null && regularNamesMap.get(otherName) !=null) {
+                    SemanticEntityVersion semanticEntityVersion = regularNamesMap.get(otherName);
+                    SemanticEntity<SemanticEntityVersion> semanticEntity = semanticEntityVersion.chronology();
+                    synonym.semantic(semanticEntity.toProxy());
+                }
+            }
```
carldea, line 346: @swaroopsalvi Can we have @dholubek weigh in on this? Things look good but I feel there's some redundancy in `OTHER_NAMES` and `REGULAR_NAMES` w.r.t the array list and map respectively.
(purely optional) - In addition, we could rename the `REGULAR_NAMES` to `OTHER_NAME_SEMANTIC_VERSION_MAP`.

swaroopsalvi: @carldea , Doug and I spoke briefly and explained him why  I implemented the way I have currently. This can be improvised for sure but then it will be better to have it as part of the bigger refactor effort since DescrName class is used at multiple places in Pattern editing.

dholubek: @carldea I talked with @swaroopsalvi and this was to avoid refactoring and breaking create pattern.  I created a story to refactor and combine the collections and use the hashmap one going forward.  
https://ikmdev.atlassian.net/browse/IIA-1166

dholubek (approved):
looks good

PR #163, opened 2024-12-11
Feature/initial general editing

carldea:
Komet now depends on a sub project dependency **knowledge layout**.


carldea:
This PR is replaced with the branch https://github.com/ikmdev/komet/pull/164 branch:`feature/initial-general-editing2`

PR #166, opened 2024-12-13
Add KLComponent, KLComponentSet and KLComponentList custom controls

jperedadnr:
This PR adds three customs controls:
- KLComponentControl
- KLComponentSetControl
- KLComponentListControl

These allow populating a single item, a set of items (without duplicates) and a list of items (with duplicates), via drag and drop.

For now, each control has its own css file, (which rely on kview.css), and a resource bundle file, but this could be changed later on, so all custom controls share the same resources.

For testing this PR, these controls can be easily be added to existing FXML files like:

```
<?import dev.ikm.komet.kview.controls.KLComponentControl?>
<?import dev.ikm.komet.kview.controls.KLComponentListControl?>
<?import dev.ikm.komet.kview.controls.KLComponentSetControl?>

...
                  <KLComponentControl title="Component"/>
                  <KLComponentSetControl title="ComponentSet"/>
                  <KLComponentListControl title="ComponentList"/>
...
```
It shows something like:
image

And after dragging and dropping some entities:
image


kview/src/main/resources/dev/ikm/komet/kview/controls/component-control.css
```diff
@@ -0,0 +1,111 @@
+@import "../mvvm/view/kview.css";
+
```
carldea, line 2: I actually didn't know about this. This is excellent!

carldea (approved):
@jperedadnr  I don't see anything glaring, except no java docs, 
Excellent work! I think this is a good first cut to have the team use these controls.

To move things along I would like to approve and merge, but Can we create another ticket to update the jdoc?

PR #168, opened 2024-12-13
Add KLByteArrayControl

jperedadnr:
This PR adds a custom control:

- KLByteArrayControl

This allows file attachments that are stored as byte array.

For now, the control has its own css file, (which rely on kview.css), and a resource bundle file, but this could be changed later on, so all custom controls share the same resources.

For testing this PR, this control can be easily be added to existing FXML files like:

```
<?import dev.ikm.komet.kview.controls.KLByteArrayControl?>
...
<KLByteArrayControl title="Files" prefWidth="300" showUploadProgress="true"/> 
...
```

Produces
image

image

image


carldea (approved):
Nice work! Nothing glaring. 
This is low priority UI control. I don't know where we use quite yet. I will approve to keep things together when we revisit this control (to be used).

Can you provide a snapshot image of the control?

PR #169, opened 2024-12-13
Add KLInstantControl

jperedadnr:
This PR adds a custom control:

- KLInstantControl

This allows selecting date and time.

For now, the control has its own css file, (which rely on kview.css), and a resource bundle file, but this could be changed later on, so all custom controls share the same resources.

For testing this PR, this control can be easily be added to existing FXML files like:

```
<?import dev.ikm.komet.kview.controls.KLInstantControl?>

...
        <KLInstantControl title="Instant:" prefWidth="300" /> 
...
```

Produces

image

image

image

image

image


carldea (approved):
Excellent work!
If possible can we try to include a screen shot image of the control?
Quite interesting way to take advantage of interrogating `DatePickerSkin`'s nodes.


I will approve. Would like another ticket to add jdocs.

PR #170, opened 2024-12-14
Feature/IIA-1170: First implementation of Sampler App

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1170

----------------------------------------------------------------------------------------------------

This is the first implementation of the Sampler App.

I believe this Sampler can add the following advantages:
1 - Implement the custom controls independently, no need to rely on anything that's not yet finished in Komet
2 - Provide an additional way to test out the custom controls... say if anything is not working, we could just test the sampler and quickly check if the custom control in itself is not working
3 - Potentially this provides a way to demonstrate or document the controls that we provide

While working on it I've discovered a few minor bugs with SortedComboBox that I also fix in this PR, so I guess one of the purposes of the Sampler is already working (testing out controls and fixing their bugs).

This Sampler also has one of the Custom Controls that was created for "Semantic Authoring" which is `KLBooleanControl`. I can also add the other 2 that were merged if you guys think it's a good idea, but I'll need to know how they work and what they're for.

The following video showcases this version of the Sampler running:

https://github.com/user-attachments/assets/63b16adf-0037-434e-bf50-039f536c4b6e


application/src/main/java/dev/ikm/komet/sampler/SamplerApp.java
```diff
@@ -0,0 +1,133 @@
+package dev.ikm.komet.sampler;
```
carldea, line 1: Can we move the Sampler Application into the test area?
`application/src/test/dev/ikm/komet/sampler/`

dukke: Done! I was having issues before on the "test" area but now have moved it to there and got it to work.

application/pom.xml
```diff
@@ -16,6 +16,10 @@
             <groupId>org.carlfx</groupId>
             <artifactId>cognitive</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.pixelduke</groupId>
+            <artifactId>fxcomponents</artifactId>
+        </dependency>
```
carldea, line 22: Can these be in **test** scope for now?

application/src/main/java/module-info.java
```diff
@@ -47,6 +47,7 @@
     requires nsmenufx;
     requires fr.brouillard.oss.cssfx;
     requires org.carlfx.cognitive;
+    requires com.pixelduke.fxcomponents;
```
carldea, line 50: Does the main project need `fxcomponents` or is this for the sampler? If it is for the Sampler we should move it into the test module.

application/pom.xml
dukke: No, the custom control is only needed in the Sampler, good point! 
Done!

dukke: > Can these be in **test** scope for now? Or does the custom control need them?

No, the custom control is only needed in the Sampler, good point!
Done!

application/src/main/java/module-info.java
dukke: Good catch. That was a mistake. Done!

dukke: > Does the main project need `fxcomponents` or is this for the sampler? If it is only for the Sampler we should remove from here and move it into the test module.

Good catch. That was a mistake. Done!

dholubek (approved):
looks good, all review comments are reconcilled.

PR #173, opened 2024-12-16
BugFix/IIA-1172: Fix scroll on next gen search in the journal view

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1172


dholubek (approved):
looks good

PR #176, opened 2024-12-18
IIA-1177: ComboBox selected value is not shown correctly

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1177


dholubek (approved):
I tested the bugfix branch across pattern and concept and everything looks good

PR #177, opened 2024-12-18
BugFix/IIA-1178: Concept Other Name Type should be non-editable and disabled

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1178


dholubek (approved):
looks good

PR #178, opened 2024-12-18
AR-251-Initial support of ObservablesAPI & KlField interfaces

carldea:
Initial support of ObservablesAPI with KlStringField, KlComponentField as editable field (widgets) in the edit bump out panel (right).


aks8m (approved):
Looks good!

PR #183, opened 2024-12-23
Feature/IIA-1180: Read only string

dukke:
**SHOULD BE MERGED ONLY AFTER https://github.com/ikmdev/komet/pull/197**

Fixes https://ikmdev.atlassian.net/browse/IIA-1180

https://github.com/user-attachments/assets/cfff2995-7063-4a34-a406-c2a9d7506030

With context menu showing:
image

PR #187, opened 2024-12-26
Feature/IIA-1181: Create read only component for component field

dukke:
**SHOULD BE MERGED ONLY AFTER https://github.com/ikmdev/komet/pull/183**

Fixes https://ikmdev.atlassian.net/browse/IIA-1181

image


carldea (approved):
Looks good, Doug and I resolved a conflict.

PR #191, opened 2024-12-31
Feature/IIA-1182: Create read only component for component set field

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1182

**THIS PR SHOULD ONLY BE MERGED AFTER THE PR FOR READ-ONLY BOOLEAN, UUID, ETC https://github.com/ikmdev/komet/pull/190**

image

image

PR #193, opened 2024-12-31
Feature/IIA-1183: Create read only component for component list field

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1183

**THIS PR SHOULD ONLY BE MERGED AFTER THE PR ABOUT COMPONENT SET FIELD (https://github.com/ikmdev/komet/pull/191)**

image


carldea (approved):
Looks good to me. We will need to revisit when we begin the work on the KlComponentField implementation to coordinate the actual data objects and the displayable text on the control.

PR #194, opened 2024-12-31
Feature/finished/ikm kec kl1

kec:
Changes to clarify use of FieldInterface and added FieldImplementation


dholubek (approved):
going to merge since FieldFactory could still be implemented and will remove those callers

PR #197, opened 2025-01-03
Add Gluon Controls to Sampler

dukke:
This PR adds the Gluon controls to the Sampler.

A println that was being done inside FloatControl has also been removed.

https://github.com/user-attachments/assets/f1ffe4f7-e33e-49a4-a914-3cff3cad19bd


dholubek (approved):
looks good

PR #199, opened 2025-01-06
Feature/ikm kec kl3

kec:
Did a merge locally, push, and here we are again.


kec:
We probably need a release of tinkar-core to eliminate the dependency on 1.80.0-SNAPSHOT

kec:
There are some binding features put in tinkar-core that the layout api requires.

PR #200, opened 2025-01-06
Feature/finished/ikm-kec-kl3

kec:
Update tinkar-core version to <tinkar-core.version>1.80.0</tinkar-core.version>


kec:
Ok, updated to the just released tinkar-core, resolved conflict's, and it seems to be building fine, and ready to merge.

PR #202, opened 2025-01-07
Feature/finished/ikm kec kl4

kec:
Added back in some things left on the cutting room floor (removal of some unused and deprecated interfaces, and binding examples)


carldea (approved):
Looks good to me!

PR #205, opened 2025-01-09
Remove textFormatter from internal textField in KLStringControl

jperedadnr:
Fixes issue that prevents typing symbols, numbers or non-alphabet characters.

The only characters not allowed (by default from the JavaFX TextField) are: Enter (c = 0xA), Tab (c = 0x9), Delete (c = 0x7F), and the ASCII control characters (c < 0x20).

The error indication doesn't make sense anymore, so it is removed as well.


bhharsh13 (approved):

dholubek (approved):
approval

PR #206, opened 2025-01-09
Remove unexpected minus handling from KLIntegerControl

jperedadnr:
This PR removes the handling of the minus sign `-` when typed in a different place other than the beginning of the integer number.


jperedadnr:
Not required anymore

PR #207, opened 2025-01-09
remove unwanted null values, remove string converter and allow for plus sign

jperedadnr:
This PR fixes a bug where null was passed to the control after a valid value was set.
It also removes the double string converter, so the number doesn't get formatted when text is committed (after losing focus). This allows undo/redo operations.
It also includes the plus sign, that is treated as the minus sign.


bhharsh13 (approved):
Looks good to me

dholubek (approved):
approved

PR #208, opened 2025-01-09
BugFix/IIA-1191: All the icons in komet app - Journal, on the left hand panel looks stretched

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1191

### Before
image

### After
image
image


kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
```diff
@@ -5015,7 +5009,15 @@ Styling a context menu for kview ui/ux controls
...
     -fx-min-height: size;
+}
 
+/*******************************************************************************
+ *                                                                             *
+ * Icons                                                                       *
+ *                                                                             *
+ ******************************************************************************/
+
+.icon {
```
dholubek, line 5020: are we sure this won't break anything?  do all the icons in Komet have a background color white?

dholubek (commented):
have a comment about using ubiquitous .icon class here.

kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
dukke: Good point. I've removed that css definition and moved it to the icon-klcontext-menu styleclass to make sure it only applies to the context menu icons

carldea: Can we add some before and after images in the PR?
I feel this fix is only localized, I thought something happened recently when we switched out something. Lot's of other buttons and icons (graphic of button/toggle, svgnode, regions, etc.) in other screens are also squished.

dukke: Images have been attached to the issue

dholubek (approved):
approved

PR #212, opened 2025-01-14
Fixes where javafx controllers using the Factory's older create methods.

carldea:
@dholubek  This is a fix to the last PR. It failed to build. We missed methods using the `boolean editable` (3rd) parameter on the factory's `create()` method.


carldea:
Quick fix to resolve an issue with the recent merge that broke the build.

aks8m (approved):

PR #230, opened 2025-01-25
IIA-1321 Implement Data Semantic Refresh

stenkarl:


carldea (approved):
Looks good to me. 

I do want us to discuss further w.r.t. the observableFields as they are set. Some are set to null. I think we can make the not null by always calling the obtainObservableField() method.

bhharsh13 (approved):
Looks good to me

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
```diff
@@ -151,14 +166,17 @@ private void clearForm(ActionEvent actionEvent) {
...
     public void submit(ActionEvent actionEvent) {
         actionEvent.consume();
-        System.out.println(actionEvent);
-        editFieldsVBox.getChildren().forEach(node -> {
-            ObservableField observableField = (ObservableField) node.getUserData();
+        List<ObservableField> list = new ArrayList<>();
+        observableFields.forEach(observableField -> {
             if (observableField != null) {
                 observableField.writeToDataBase();
             }
```
dukke, line 173: Shouldn't we be calling writeToDataBase on each observableField as well?

PR #231, opened 2025-01-28
Feature/finished/IIA-1326: Implement Dropdown (Combo Box) for Editable Boolean Control

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1326

-------------------------

### Changes in this PR
1. Creates a complete whole new editable Boolean Control (Gluon's control was tied to being a radio button)
2. Tweak to read-only data field control so that the first letter is always capitalized

### Notes
1.  Had to put some of the CSS code in KView namely for the combobox popup as it wasn't working when I put in the control css file. This could be happening for 2 reasons (which I didn't have time to explore): 
1.1 - Something in kview was overriding my css for the popup (control's css has less priority than kview.css because kview.css is an author stylesheet
1.2 - The popup is in a whole new scene (it is a stage that is created when the combox is open)

### Screenshot
image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.46.0-SNAPSHOT</version>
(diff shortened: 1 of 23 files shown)


dholubek (approved):
looks good

PR #247, opened 2025-02-05
Feature/finished/iia 1112 image data type

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1112

This PR adds in Image Data type in Komet. The image data type is hardcode, nothing is fetched from the database as we don't have that capability yet in the DB.
For now, there was no need to restrict the type of images that can be added but the file dialog does restrict the users choice to only images.

It will only show in one of the test semantics. 

I've attached a video to the Jira ticket given it was too big for Github. 

**Update 06/02/2025: Image file name UI has been removed from the UI in light of today's discussion with Amplified and PR has been updated accordingly**

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/klfields/imagefield/KlImageFieldFactory.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/klfields/imagefield/KlImageFieldFactory.java
@@ -31,6 +31,6 @@ public class KlImageFieldFactory implements KlFieldFactory<Image> {
 
     @Override
     public String getDescription() {
-        return "A Image field";
+        return "An Image field";
     }
 }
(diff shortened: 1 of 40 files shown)


kview/src/main/java/dev/ikm/komet/kview/klfields/imagefield/KlImageFieldFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "Image Field Factory";
+    }
+
+    @Override
+    public String getDescription() {
+        return "A Image field";
```
stenkarl, line 34: A super-small nit, but grammatically this would be "An Image Field"

dukke: Good catch. Fixed.

stenkarl (approved):

PR #251, opened 2025-02-08
Feature/finished/ikm kec kl5

kec:
Knowledge layout model evolution.


kec:
Created KL6 pull request to replace this KL5 pull request.

PR #256, opened 2025-02-11
Feature/finished/ikm kec kl6

kec:
New pull request to remove reference to google CaseFormat reference.


carldea (approved):
@kec Looks good to me! 
Nice documentation (javadocs)! 

I did noticed your TODO related to Java 21's `Subscriber` api as opposed to the older property change listeners. We are tracking this and plan to use it more often especially situations where memory leaks could occur.
  
Thank you for refactoring the `KlWindow` code. I look forward to more integration with the new interfaces especially the preferences functionality.

PR #258, opened 2025-02-11
Feature/finished/iia 1328 read only component set control

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1328

This PR adds a Read Only Component Set Control and uses it in the Semantic Authoring Left Panel. 

Tried to make sure everything is "pixel perfect".

Also fixes:
1.  Make spacing consistent in the various controls: Read Only Component Control and Read Only Component Set Control
2. Make hover and edit mode for Read Only Component Control look like in the Figma


kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLReadOnlyComponentSetControlSkin.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    /*******************************************************************************
+     *                                                                             *
+     * Supporting Classes                                                          *
+     *                                                                             *
+     ******************************************************************************/
+
+    public static class ComponentItemNode extends Region {
```
carldea, line 122: Could this class be pulled out so it can also be used in Component List?

carldea: Or are they different. E.g. Set does not have (numbers/index) beside them like for Component List.

stenkarl (approved):

PR #261, opened 2025-02-12
Feature/finished/iia 1456: Exception Triggered when we click on "Tinkar semantic test pattern" in Pattern Navigator

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1456


carldea (approved):
Looks good!

PR #267, opened 2025-02-13
Feature/finished/iia 1273 semantic purpose and meaning

bhharsh13:
Fixed the Semantic meaning and Semantic purpose text overlap issue.

https://ikmdev.atlassian.net/browse/IIA-1273

Validated the changes locally. Please help to review the PR.

Before resizing the Pattern
Before resizing Pattern
After resizing the Pattern, text doesn't overlap for both.
After resizing Pattern - no text overlap

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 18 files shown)


stenkarl (approved):

carldea (approved):
Great job! I like how both columns are responsive and honors the parent's available width. Mastering layout is hard and is always different depending on the scenario.

PR #269, opened 2025-02-14
Feature/finished/iia 1446 image database support

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1446

Features implemented in this PR:
1. Uses ByteArray type so no longer relies on nids (which means it will now work on any machine not just mine)
2. read image from database 
3. sync left read only pane once submit is clicked
4. save back to database once submit is clicked

Caveats:
- Currently we support jpg, jpeg, gif and BMP
- Images are always stored as PNGs in the database


kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -96,12 +92,14 @@ public static List<ObservableField<?>> displayEditableSemanticFields(ViewPropert
             } else if (dataTypeNid == TinkarTerm.BOOLEAN_FIELD.nid()) {
                 KlBooleanFieldFactory klBooleanFieldFactory = new KlBooleanFieldFactory();
                 node = klBooleanFieldFactory.create(observableField, viewProperties.nodeView(), true).klWidget();
+//            } else if (dataTypeNid == TinkarTerm.IMAGE_FIELD.nid()) {
```
stenkarl, line 95: Do you want to remove the commented lines?

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -158,48 +149,19 @@ public static List<ObservableField<?>> displayReadOnlySemanticFields(ViewPropert
                 } else if (dataTypeNid == TinkarTerm.BOOLEAN_FIELD.nid()) {
                     KlBooleanFieldFactory klBooleanFieldFactory = new KlBooleanFieldFactory();
                     readOnlyNode = klBooleanFieldFactory.create(observableField, viewProperties.nodeView(), false).klWidget();
+//                } else if (dataTypeNid == TinkarTerm.IMAGE_FIELD.nid()) {
```
stenkarl, line 152: Do you want to remove these commented lines?

carldea: I thought the database field type would be recognized as an IMAGE_FIELD type. but read and written as a byte array? Byte array field is a different field type (that also read and writes as a byte array)

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -158,48 +149,19 @@ public static List<ObservableField<?>> displayReadOnlySemanticFields(ViewPropert
...
+//                    KlImageFieldFactory imageFieldFactory = new KlImageFieldFactory();
+//                    readOnlyNode = imageFieldFactory.create(observableField, viewProperties.nodeView(), false).klWidget();
+                } else if (dataTypeNid == TinkarTerm.BYTE_ARRAY_FIELD.nid()) {
+                    KlImageFieldFactory imageFieldFactory = new KlImageFieldFactory();
```
carldea, line 156: We would not support currently, but we should create a read-only custom control and an editable control for raw binary data. Possibly showing a preview of bytes in hex. but this will need to be prioritized and worked on later.

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
carldea: I didn't see this until now. I added comments in the other `else if` code for the `displayReadOnlySemanticFields()` method below.

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
stenkarl: The change for the image data type to be recognized as its read from the database is invasive enough where it needs some eyes from Keith. Since we don't use the byte_array type for anything else we can assume it to be an image for now.

dukke: @stenkarl I left that comment there for later so that when we have IMAGE data field type we can uncomment. Do you prefer to not have it?
@carldea we don't have support for IMAGE data field type yet. When we do, we can return here and change the if condition

stenkarl: It's fine, you can leave it.

stenkarl (approved):

PR #276, opened 2025-02-19
Feature/finished/iia 1463 component set ordering

bhharsh13:
https://ikmdev.atlassian.net/browse/IIA-1463

We were able to fix the issue when we add or delete the pattern,  the Set is ordered in sync with the left and right panels. We have an edge case for the below scenario:

“If we add or delete the pattern from top or in between , set ordering in the left is not in sync with right panel.”
This is an edge case that we will address in next sprint.

Validated the flows on local.  PFA
image

Before Editing the pattern on the right.
image

Delete or add the patterns and click Submit. The ordering for the Set is in sync with both the panels.
image

After submit
image

Add new pattern
image

We have an edge case for this scenario- “If we add or delete the pattern from top or in between , set ordering in the left is not in sync with right panel.” This is an edge case that we will address in next sprint.
image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.47.0-SNAPSHOT</version>
+        <version>1.48.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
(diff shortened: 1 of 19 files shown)


carldea (approved):
Looks good @bhharsh13 . Thank you for creating another ticket to solve the other scenario (removing one and reordering). We can solve in the next sprint.

PR #277, opened 2025-02-20
Feature/finished/iia 1481 six dots icon not clickable

bhharsh13:
Fixed the issue for Pattern Tile Six-Dots Icon to not to Be Clickable.
Moved the Clickable Menu to appear on Semantic Row.
Please help to review the PR.
Validated the changes in local.
image
image


kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/navigation/pattern-nav-entry.fxml
```diff
@@ -35,29 +32,7 @@
...
-                                    <Region styleClass="remove-icon">
-                                       <padding>
-                                          <Insets right="16.0" />
-                                       </padding>
-                                    </Region>
-                                 </graphic>
-                              </MenuItem>
-                           </items>
-                        </ContextMenu>
+                        <ContextMenu fx:id="contextMenu" styleClass="search-context-menu-container" />
```
carldea, line 35: Can we remove line 35 and the surrounding `</contextMenu>`?

carldea: @bhharsh13 Can we also have @dukke review to make sure I'm not missing something?

Thank you! 👍🏼

dukke (commented):
(Sorry I added a comment above but not as a review comment, adding it here again)

Hey @bhharsh13 thanks for this.

I think the ticket description might be wrong.

I see that the context menu's contents have been removed, is that right? And that fulfils what's wanted in the ticket. But I think the actual idea is not to remove the context menu entirely, so I think the description in the ticket is wrong.

We don't want to remove the contents of the context menu but rather to fix the issue we want the context menu to open on right click on the whole Pattern instead of just opening when you left or right click the six dots.

Could you also possibly add a link to the ticket?

Any thoughts @carldea ?

Thanks!

kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/navigation/pattern-nav-entry.fxml
carldea: @bhharsh13  I agree with what @dukke is suggesting.

In other words, move the context menu (and options) for the **Pattern row** and not the six dots. I believe this ticket stating to remove options. We want the context menu to work just not on the six dots icon.

We should update the ticket story and acceptance criteria to reflect this. If there is already a ticket It should be combined. 
**A possible user story title:**
As a user, i can perform a context menu popup for the Pattern (row) as a whole and not on the six buttons.

dukke (commented):
I think the context menu needs to be set on another place. Thanks!

PR #279, opened 2025-02-21
Feature/iia 1464 debouncing poc try2

dholubek:
This is a debouncing proof of concept in the test area of application


carldea (approved):
Looks great to me. I think this should be good enough for our needs.

PR #280, opened 2025-02-22
Feature/finished/iia 1335 numbered editable list

dukke:
### Fixes:
- https://ikmdev.atlassian.net/browse/IIA-1462
- https://ikmdev.atlassian.net/browse/IIA-1395
- https://ikmdev.atlassian.net/browse/IIA-1335

### Video capture of how it's working with this PR:
https://github.com/user-attachments/assets/704ed104-fe85-4dd5-94df-644d39b41453


carldea:
@dukke Pedro, 
Really glad you have reordering with a green line working! 
But I think there is a subtle difference in the way the figma spec. describes the scenarios. Also think the specification changed recently after your PR.

Let's confirm this together & design team. I was thinking of two option for this PR:
1. We could merge your work (good enough for now) and create another ticket to make the changes based on the newer specs.
2. We could continue to work on this PR (Put in draft) to follow the specs.

**Notes:**
Scenarios when reordering: ( A caveat - I could be completely wrong 👎🏼 : -)
1. Last item to the first item - green line needs to be above the first item (appearing as an insert) pushing items down.
2. First item or outside item dragged to the bottom. - green line needs to be below the last item pushing items up.
Screenshot 2025-02-23 at 5 05 33 PM
The above example shows the 5th item dragged to be the 2nd item.

**Outstanding questions:**
1. Based on the above we should have more Figma example wireframes covering all scenarios.
2. The figma specs shows a white blank background when dragging the item back in place.
Screenshot 2025-02-23 at 5 09 40 PM
3. We now have an addition 'X' beside the new item from the Add Entry button. Can we just have one?
Screenshot 2025-02-23 at 5 13 50 PM

4. In the figma specs why is the user allowed to add more than one blank field.
Screenshot 2025-02-23 at 5 02 37 PM

dukke:
@carldea Thanks for the detailed comments to this PR.

### Notes:

1 - I think I understand what you're saying but there are practical reasons to why that doesn't happen on this PR. We don't have space between the title and the first component to draw a drop line there. We can add more space between the title and the first component but then we'll have to do that for every editable component for consistency's sake. Perhaps we could do that, but we might want to add another ticket as this one is already getting too big (it's already fixing 3 issues) and contains tickets there were target for the previous sprint.

2 - Not sure I understand this one but maybe we can talk in a video meeting...

### Outstanding questions:

Those are good questions but maybe we can tackle them outside of this PR as this PR is already too big and also references tickets that were already in a previous sprint? So perhaps we finish or merge this (if we agree on how things are implemented and done) and then we can tackle other issues on a different ticket?

carldea:
I agree it is getting large (taking too long) and changes in the Figma spec happened after your merge was created. 

At least we have a working version. 👍🏻

carldea (approved):

PR #282, opened 2025-02-26
Feature/finished/iia 1474 code cleanup

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1474

-----------------------------

1. removed old classes that are no longer being used or needed
2. deprecated ReadOnlyKLFieldFactory. We're only using it to display DiTree and will be removed once we have a proper implementation of that
3. moved component set "field classes" to componentset package

Recompiled everything from scratch to make sure there were no compilation errors and tested Komet Semantic editing.


kview/src/main/java/dev/ikm/komet/kview/klfields/readonly/ReadOnlyKLFieldFactory.java
```diff
@@ -60,93 +48,93 @@ public Node createReadOnlyDiTree(ViewProperties viewProperties, FieldRecord<?> f
...
+//        IntIdSet componentSet = (IntIdSet) fieldRecord.value();
+//        componentSet.forEach(componentId -> {
+//            Latest<EntityVersion> component = viewProperties.calculator().stampCalculator().latest(componentId);
+//            component.ifPresent(entityVersion -> {
+//                Node componentRow2 = createReadOnlyComponentListItem(viewProperties, component);
+//                container.getChildren().add(componentRow2);
+//            });
+//        });
+//        return componentRow;
+//    }
```
carldea, line 95: I think we should remove these commented out functions. :-)
Good call btw, regrading DiTree and DiGraph. Once we create these controls we can remove class.

dukke: Agree, it's better and cleaner if we just remove the commented-out code. 
Just did that and pushed.

carldea (approved):
Great work. Nice clean up.

We should test to make sure the digraph and ditree still outputs. It's still prototype code until we get that in place.

PR #288, opened 2025-02-28
Feature/finished/iia 1474 code cleanup

dukke:
fixes https://ikmdev.atlassian.net/browse/IIA-1474


carldea (approved):
Things look great. I assume this is the same as the previous PR.

PR #295, opened 2025-03-03
Feature/finished/iia 1459 pattern navigator redirect

dholubek:

right click from next gen nav opens concept (which it already did behind the scenes but the pattern toggle is the default) but it toggles the concept toggle so that the concepts are showing.


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
```diff
@@ -388,6 +391,10 @@ public void initialize() {
         workspace.addEventHandler(MouseEvent.MOUSE_CLICKED, this::onMouseClickedOnDesktopSurfacePane);
     }
 
+    private void toggleConcepts() {
+        conceptPatternNavController.toggleConcepts();
```
carldea, line 395: Can this be called in line 378 instead of a function?

Since it is only one line statement creating a method as passthrough method can add a layer of indirection.

dholubek: I can try but it might be a timing issue when the subscriber's lambda is initialized and when the config make() is called for this controller...

dholubek: fixed

carldea (approved):
Looks good to me!

dholubek:
can't be rebased safely... so closing

PR #296, opened 2025-03-03
Feature/finished/iia 1317 make scrollbar thinner

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1317

image


carldea (approved):
While it appears fine. I noticed the width of the content is not taking the available width.
Screenshot 2025-03-03 at 4 14 28 PM

Can you have @dukke examine it?

dukke:
> While it appears fine. I noticed the width of the content is not taking the available width. Screenshot 2025-03-03 at 4 14 28 PM
> 
> Can you have @dukke examine it?

@carldea that was there before this PR. Might be that the background colors of the container and the container of the container aren't the same.

dukke:
@dholubek @carldea I see that in the screenshot the read only reference component is being cut out... that didn't happen before. Is it something this PR did or more likely some other PR that has been recently merged? We should file a ticket to fix it, I think.

dholubek:
> @dholubek @carldea I see that in the screenshot the read only reference component is being cut out... that didn't happen before. Is it something this PR did or more likely some other PR that has been recently merged? We should file a ticket to fix it, I think.
> 
> Update: actually, no PR got merged recently so maybe it's this PR or some edge case that makes the reference component get cut off.

Per chat:
1) ref component = existing and separate bug ticket
2) semantic details doesn't fill up the full area (and is not white for a background like pattern) = existing and separate bug ticket
3) image control and klFieldHelper = zombie defect that reemerged that @dukke  is re-applying the fix for

dholubek:
closing since rebase is not working for this merge

PR #298, opened 2025-03-03
Feature/finished/iia 1546 image field exception

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1546

image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.48.0-IIA-1546-Image-Field-Exception-SNAPSHOT</version>
+        <version>1.48.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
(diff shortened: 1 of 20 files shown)


dukke:
@carldea 
Something happened to Tinkar I think. A Tinkar Image/Bytearray nid that used to work no longer works. I came up with a workaround (using the PublicID class), but we might need to warn the Tinkar team.

dukke:
> @carldea Something happened to Tinkar I think. A Tinkar Image/Bytearray nid that used to work no longer works. I came up with a workaround (using the PublicID class), but we might need to warn the Tinkar team.

@carldea Might also be our starter test data that needs to be updated. The image that we get from there might not have the correct id.

carldea:
> > @carldea Something happened to Tinkar I think. A Tinkar Image/Bytearray nid that used to work no longer works. I came up with a workaround (using the PublicID class), but we might need to warn the Tinkar team.
> 
> @carldea Might also be our starter test data that needs to be updated. The image that we get from there might not have the correct id.

Agreed. We need to verify this with the Data team.

dholubek (approved):
looks good

carldea (approved):
Looks good to me! Thanks.

PR #300, opened 2025-03-04
Feature/finished/jpro 103 concept opens in next drop area on duplicate entry error

indritbeqiri:
**JIRA Ticket**: [JPRO-103](https://ikmdev.atlassian.net/browse/JPRO-103)

**Details**: Related before and after videos are attached to the Jira ticket.

[JPRO-103]: https://ikmdev.atlassian.net/browse/JPRO-103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


indritbeqiri:
@carldea 
On Windows, we observed that occasionally the drop could still be processed even if a duplicate was detected during the drag over event. To address this, explicitly calling setDropCompleted(false) in the drag dropped handler was necessary to prevent the duplicate entry from being processed.

PR #301, opened 2025-03-04
IIA-1459 show concept in navigator

dholubek:

right click from next gen nav opens concept (which it already did behind the scenes but the pattern toggle is the default) but it toggles the concept toggle so that the concepts are showing.


carldea (approved):
Looks good to me for the 3 or 4th time.

PR #302, opened 2025-03-04
Feature/finished/iia 1559 ref comp cut off

dholubek:
DRAFT: the thinner scroll bar somehow got reverted... will fix this and update this PR...

https://ikmdev.atlassian.net/browse/IIA-1559

image

image


dukke:
@dholubek @carldea the main issue here I think is the read only control itself reporting a wrong min height. We shouldn't need to set a minimum height in the pane.
So, I think we should rather fix the min height that the control reports, otherwise we'll need to keep setting min height every time there's a change in the pane or on any place where we have the read only component control. 
There's a method you need to override for that, that method is computeMinHeight in the skin of the control.

dholubek:
this is missing the thin scrollbar

PR #304, opened 2025-03-05
Feature/finished/iia-1558 duplicate concept list & IIA-1535 reorder bug

dukke:
fixes https://ikmdev.atlassian.net/browse/IIA-1558 and https://ikmdev.atlassian.net/browse/IIA-1535

- side note: also fixes a small UI glitch in Pattern Browser when a semantic is clicked


carldea (approved):
Excellent. Glad this is working. Lists can be tricky.

PR #305, opened 2025-03-05
Feature/finished/iia 1459 fix

dholubek:

fix where the toggle buttons needed change listeners if we programmatically fired the toggle

image

image


bhharsh13 (approved):
Looks good to me

PR #306, opened 2025-03-05
Feature/finished/iia 1559 pattern ref comp height

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1559

before:
image

after:
image


carldea:
Can you add some before and after screen shots of the UI?

carldea (approved):
I believe this is an okay fix.

PR #307, opened 2025-03-05
Feature/finished/iia 1560 pattern sem window color consistency

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1560

image


dukke:
@dholubek @carldea The background is not white and should be white but I've been hit with this before where even though I set it to white it shows up as gray, it's weird, not yet sure what's going on (but if you set it to red, green or blue it does look that color). And also, we can set it to White in the Concept Window and it works.

Have you tried setting it to something real close to white like: #fefefe ?

carldea:
> you tried setting it to something real close to white like: #fefefe ?

That's a great idea. It's unusual if it doesn't like `#ffffff`.

dholubek:
> @dholubek @carldea The background is not white and should be white but I've been hit with this before where even though I set it to white it shows up as gray, it's weird, not yet sure what's going on (but if you set it to red, green or blue it does look that color). And also, we can set it to White in the Concept Window and it works.
> 
> Have you tried setting it to something real close to white like: #fefefe ?

I'll give that a try

dukke:
> > @dholubek @carldea The background is not white and should be white but I've been hit with this before where even though I set it to white it shows up as gray, it's weird, not yet sure what's going on (but if you set it to red, green or blue it does look that color). And also, we can set it to White in the Concept Window and it works.
> > Have you tried setting it to something real close to white like: #fefefe ?
> 
> I'll give that a try... this is what #fefefe looks like: image

Thanks @dholubek, good work!
Still looks grey, there's something weird going on, not sure what could be. Like I said I've also experienced this issue before but also didn't came to any conclusion back then.
One thing could be some semitransparent thing on top that makes the colors different, that we see in the end on the screen.

Not sure if you guys @dholubek @carldea want to merge it as is, set the color in CSS to white, leave a comment there saying "TODO: for some reason this is not setting the color to white" and create a story to track that problem?

Thanks!

carldea (approved):
Thanks for examining this issue. I will approve and we can make another ticket to come back to this.

PR #312, opened 2025-03-10
Feature/finished/iia 1279 score order search matching

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1279

The score order search result did not match exactly from Next Gen Search and Classic Search
The classic search uses a TreeItem in the populateTempRoot() method.  The next gen search loads an FXML for the result entries and used a HashMap to hold them, that map was unordered and changing it to a LinkedHashMap to preserve insertion order allowed the logic to match classic Komet

BEFORE:
before-out of order

AFTER:
after-matching

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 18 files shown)


swaroopsalvi (approved):
Approve

PR #316, opened 2025-03-11
Feature/finished/iia 1458 disable enable submit button when editing semantic

swaroopsalvi:


carldea (approved):
Looks good to me!
I'm assuming after the submit pressed there is addition clean up code. e.g. after submit the new committedHash becomes the new version.

PR #318, opened 2025-03-11
Feature/finished/iia 1594 component list reorder dropline bug

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1594


carldea (approved):
Thanks! Looks good to me.

PR #319, opened 2025-03-12
Feature/finished/iia 1550 show 6 dots on hover concept

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1550

image


bhharsh13 (approved):
It would be good if we can attach the screenshot validated local when creating the PR. Looks good to me.

dukke:
> It would be good if we can attach the screenshot validated local when creating the PR. Looks good to me.

Good idea. Image attached.

PR #323, opened 2025-03-14
Feature/finished/iia 1602 write to database and autosave temp bug fix

swaroopsalvi:
This is a temp bug fix for wtireToDB / AutoSave.


kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -108,17 +116,62 @@ public static List<ObservableField<?>> generateObservableFieldsAndNodes(ViewProp
...
+        observableSemanticVersionOptional.ifPresentOrElse( (p) -> {
+            entityVersionLatest.set(new Latest<>(p));
+        }, () -> {entityVersionLatest.set(semanticEntityVersionLatest);});
+
+
+        return entityVersionLatest.get();
+    }
+
+    public static int calculteHashValue(List<ObservableField<?>> observableFieldsList ) {
+        StringBuilder stringBuilder = new StringBuilder();
```
dholubek, line 171: can we have a brief javadoc for this method?

dholubek (commented):
had a couple comments for adding javadocs to public methods

kview/src/main/java/dev/ikm/komet/kview/mvvm/model/DataModelHelper.java
```diff
@@ -308,12 +309,36 @@ public static List<FieldRecord<Object>> fieldRecords(SemanticEntityVersion seman
...
      */
-    public static ObservableField<?> obtainObservableField(ViewProperties viewProperties, Latest<SemanticEntityVersion> semanticEntityVersionLatest, FieldRecord<Object> fieldRecord){
+    public static ObservableField<?> obtainObservableField(ViewProperties viewProperties, Latest<SemanticEntityVersion> semanticEntityVersionLatest, FieldRecord<Object> fieldRecord, boolean uncommitted){
         ObservableSemantic observableSemantic = ObservableEntity.get(semanticEntityVersionLatest.get().nid());
         ObservableSemanticSnapshot observableSemanticSnapshot = observableSemantic.getSnapshot(viewProperties.calculator());
-        ImmutableList<ObservableField> observableFields = observableSemanticSnapshot.getLatestFields().get();
-        return observableFields.get(fieldRecord.fieldIndex());
+        ImmutableList<ObservableSemanticVersion> observableSemanticVersionImmutableList = observableSemanticSnapshot.getHistoricVersions();
+        if(uncommitted || observableSemanticVersionImmutableList == null || observableSemanticVersionImmutableList.isEmpty()){
+            //Get the latest version which is uncommited version
```
dholubek, line 319: can we put spaces before after the if condition and the else keyword like this:

`if (uncommited) {
} else {`

instead of 
`if(uncommited){
}else{`

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -108,17 +116,62 @@ public static List<ObservableField<?>> generateObservableFieldsAndNodes(ViewProp
...
+
+        observableSemanticVersionOptional.ifPresentOrElse( (p) -> {
+            entityVersionLatest.set(new Latest<>(p));
+        }, () -> {entityVersionLatest.set(semanticEntityVersionLatest);});
+
+
+        return entityVersionLatest.get();
+    }
+
+    public static int calculteHashValue(List<ObservableField<?>> observableFieldsList ) {
```
carldea, line 170: Can we also include the Semantic's reference component's `publicId`?
My reasoning: During create (new Semantic) the user can change the reference component. the hash is stating any change. By default a reference component during created would be `TinkarTerms.ANONOUMOUS_CONCEPT` (I can't remember).

framework/src/main/java/dev/ikm/komet/framework/observable/ObservableField.java
```diff
@@ -70,6 +70,7 @@ private void handleValueChange(Object newValue) {
 
     public void writeToDataBase() {
         this.writeToDatabase(value());
+        fieldProperty.set(field().withValue(value()));
```
dukke, line 73: Question: do we want to use field or value? Not sure but we should be mindful that there seems to be a sync issue between field and value

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -108,17 +116,62 @@ public static List<ObservableField<?>> generateObservableFieldsAndNodes(ViewProp
...
+        StringBuilder stringBuilder = new StringBuilder();
+        observableFieldsList.forEach(observableField -> {
+            stringBuilder.append(observableField.valueProperty().get().toString()).append("|");
```
dukke, line 173: Maybe out of scope for this issue but we might want to be careful and not just use toString for complex values like IntIdSet and IntIdList but actually iterate through those collections (when that's the case) and do toString on each item. toString for those collections are usually really meant for debugging purposes, etc and not as a reliable source of how whether the list or set was changed. Example: a toString for a Set or List might return the same value for 2 different Sets or Lists when the only difference is the order that they have. 
Bottomline toString for a complex object (IntIdSet or IntIdList) is not reliable as a way to know exactly what they contain. But again might be out of scope for this PR.

carldea (approved):
Looks good to me. Thanks @swaroopsalvi !

PR #326, opened 2025-03-19
Feature/finished/fixed null pointer for submit button enable disable

swaroopsalvi:


carldea (approved):
@swaroopsalvi Good find. Glad we are using subscribe() more (instead of addListener). Thank you!

PR #328, opened 2025-03-19
Feature/ikm 20250319 rework reasoner flow

emays:


classification/src/main/java/dev/ikm/komet/reasoner/ReasonerResultsNode.java
```diff
@@ -113,26 +115,29 @@ public ReasonerResultsNode(ViewProperties viewProperties, KometPreferences nodeP
 				LOG.info("Default ReasonerService: " + this.reasonerService.getName());
 				menuItems.add(new SeparatorMenuItem());
 				{
-					MenuItem item = new MenuItem("Run full reasoner");
-					item.setOnAction(ae -> {
-						runFullReasoner();
+					MenuItem item = new MenuItem("Run reasoner");
+					item.setOnAction(_ -> {
+						runReasoner();
```
carldea, line 120: @emays Out of curiosity could the UI just have a single menu option to Run Reasoner?
The thought was if the system knew it was previously run, couldn't it just run the incremental reasoner?
Does the user really need to know?

carldea (approved):
Looks good to me. Glad the user has one option to run the reasoner (as opposed to two).

PR #332, opened 2025-03-20
Feature/finished/iia 1648 data on the left side of semantic window reverts back to its previous value after user close and reopen the window

swaroopsalvi:
This fix is a better approach than the previous one since it shows if the transaction exists for the current submit request as well as pulls in the latest transaction more accurately.


carldea (approved):
Looks good! We'll revisit when we resolve survive reboot and autosave.

PR #333, opened 2025-03-20
Feature/finished/iia 1471 component default value

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-1471

_Sidenote: Also made some small UI tweaks (improvements) to the left side_

image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.49.0-SNAPSHOT</version>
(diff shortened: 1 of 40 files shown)


carldea (approved):
This looks good to me. Excellent. Thank you for temporarily commenting out the disabling of the submit button. We will keep an eye on the validation code when auto-save becomes available.

To recap:

1. When user changes field (dnd and then remove) an uncommitted record (version) is saved (to the database) containing the `EMPTY_NID` value. And submit button will be disabled.
2. When a submit button is enabled the component field will contain a valid Concept nid. After user presses submit a committed record (version) is saved (to the database).

PR #334, opened 2025-03-21
Feature/finished/iia 1417 new semantic from pattern window

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1417

image


swaroopsalvi (approved):
Looks good.

PR #335, opened 2025-03-21
IIA-1484 Create Properties Panel Page to Add Reference Component When Creating Semantic

bhharsh13:
Implemented Create Properties Panel Page to Add Reference Component When Creating Semantic
https://ikmdev.atlassian.net/browse/IIA-1484

1. Given that I am creating a new semantic, when I click the Reference Component pencil icon, the Properties Panel should be opened (focused on Add/Edit) and I should see the label “Reference Component” with an empty search field.
2.Given that I see the empty search field in the Properties Panel, I should be able to drag/drop the concept I want to add into the Search field.

Regarding save implementation we are checking with Andrew in Architect hours meet.

Validated the changes on my local and the changes looks good.
image
image
image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
@@ -18,8 +18,7 @@ package dev.ikm.komet.kview.mvvm.view.genediting;
 
 import static dev.ikm.komet.kview.mvvm.viewmodel.FormViewModel.CURRENT_JOURNAL_WINDOW_TOPIC;
 import static dev.ikm.komet.kview.mvvm.viewmodel.FormViewModel.VIEW_PROPERTIES;
-import static dev.ikm.komet.kview.mvvm.viewmodel.GenEditingViewModel.SEMANTIC;
-import static dev.ikm.komet.kview.mvvm.viewmodel.GenEditingViewModel.WINDOW_TOPIC;
+import static dev.ikm.komet.kview.mvvm.viewmodel.GenEditingViewModel.*;
 import static dev.ikm.tinkar.provider.search.Indexer.FIELD_INDEX;
 import dev.ikm.komet.framework.events.EvtBusFactory;
 import dev.ikm.komet.framework.events.Subscriber;
@@ -73,7 +72,7 @@ public class PropertiesController {
 
     private Pane closePropsPane;
 
-    private ClosePropertiesController ClosePropertiesController;
+    private ClosePropertiesController closePropertiesController;
 
     @InjectViewModel
     private SimpleViewModel propertiesViewModel;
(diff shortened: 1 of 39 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
```diff
@@ -84,12 +83,54 @@ public enum PaneProperties {
...
+//                .addNamedViewModel(new NamedVm("propertiesViewModel", propertiesViewModel));
+//
+//        JFXNode<Pane, ClosePropertiesController> closePropsJfxNode = FXMLMvvmLoader.make(closePropertiesConfig);
+
+//        closePropsPane = closePropsJfxNode.node();
+//        ClosePropertiesController = closePropsJfxNode.controller();
+//        genEditingEventSubscriber = evt -> {
+//            LOG.info("Publish event type: " + evt.getEventType());
+//            contentBorderPane.setCenter(closePropsJfxNode.node());
+//        };
```
swaroopsalvi, line 118: Can we remove the commented code?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/ReferenceComponentController.java
```diff
@@ -0,0 +1,80 @@
...
+        // clear all reference details.
+        referenceComponentVBox.setSpacing(8.0);
+        referenceComponentVBox.getChildren().clear();
+        submitButton.setDisable(false);
+        EntityFacade refComponent = referenceComponentViewModel.getPropertyValue(REF_COMPONENT);
+        KLComponentControl componentControl = new KLComponentControl();
+        componentControl.setTitle("Reference component");
+//        // entity
+//        EntityProxy entity = field().value();
+//        componentControl.setEntity(entity);
```
swaroopsalvi, line 48: Remove commented code.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
```diff
@@ -27,6 +27,7 @@
 import dev.ikm.komet.framework.events.EvtBusFactory;
 import dev.ikm.komet.framework.observable.ObservableField;
 import dev.ikm.komet.framework.view.ViewProperties;
+import dev.ikm.komet.kview.controls.KLComponentControl;
```
swaroopsalvi, line 30: This class is not changed? if not changed can we remove it from this PR?

swaroopsalvi (commented):
Have few minor comments.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
bhharsh13: removed the commented code.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
bhharsh13: Removed the unused import which reflected changes in the SemanticFieldsController

dholubek (dismissed):
@bhharsh13 we should fix the component text issue before submitting a PR

dholubek:
> @bhharsh13 we should fix the component text issue before submitting a PR

original screenshot had the wrong component with drag and drop.  this is working

dholubek:
> Nice work @bhharsh13 ! On the first screen it's showing a Concept "Transitive Feature" on the left and nothing on the right... @dholubek @swaroopsalvi @carldea correct me if I'm wrong but I think in principle that shouldn't be possible (once a reference component has been set it can't be changed) or is that going to be handled in a different ticket.
> 
> One other minor comment is: would it be possible to show the part of the screen that has the buttons at the bottom (I suppose you're not able to show the whole screen because your monitor is too small?)

@dukke this ticket doesn't handle save.  So I think the limitation of one reference component post save will be handled after save is handled.

swaroopsalvi (commented):
@bhharsh13 , Just have few questions regarding the screenshots, 

1. I am sure you tested it but how come I see 

image

Transitive feature on left and blank on the right, shouldn't it should transitive feature on the left ? Unless the value on the left says place holder, the value on the right should reflect the value on the left. This is true in case of Creating semantic only, else the user should not be able to edit the reference component. 

2.   Another observation is that the 2nd screenshot STAMP does not match with the 1st one. 

image

swaroopsalvi:
> > Nice work @bhharsh13 ! On the first screen it's showing a Concept "Transitive Feature" on the left and nothing on the right... @dholubek @swaroopsalvi @carldea correct me if I'm wrong but I think in principle that shouldn't be possible (once a reference component has been set it can't be changed) or is that going to be handled in a different ticket.
> > One other minor comment is: would it be possible to show the part of the screen that has the buttons at the bottom (I suppose you're not able to show the whole screen because your monitor is too small?)
> 
> @dukke this ticket doesn't handle save. So I think the limitation of one reference component post save will be handled after save is handled.

@dukke , IMO the suer should be able to edit the reference component in create mode, even if it is submitted ? So I am in create mode and I create the reference component for the semantic, then I want to change it since I realized that I made a mistake in creation (I have not submitted/saved) the data. But even If I have have submitted the data and not published it yet, I should be able to edit it. 
  This is something we should ask Functional lead or Keith.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/ReferenceComponentController.java
bhharsh13: Removed the commented code

bhharsh13:
Hi @dukke ,
We have defect to make the edit reference component to be disabled in next sprint. The current functionality is create mode for adding new reference component. In the current defect, on clicking the edit function we are opening the property bump out adding new reference component and drag drop the reference component.

dholubek:
> @swaroopsalvi , The screenshots attached were from different timings. We have defect to make the edit reference component to be disabled in next sprint. The current functionality is create mode for adding new reference component. If you see the Acceptance criteria for the defect- In this defect we are adding a field for reference component.
> 
> 1. Given that I am creating a new semantic, when I click the Reference Component pencil icon, the Properties Panel should be opened (focused on Add/Edit) and I should see the label “Reference Component” with an empty search field.
> 2. Given that I see the empty search field in the Properties Panel, I should be able to drag/drop the concept I want to add into the Search field.
> 
> updated the screenshots in the PR. Please check if this looks good.

@bhharsh13 @swaroopsalvi  for 1) we should ask at the end of stand up or in team chat if the reference component [place holder] on the left for new semantic is in scope for this story.  I see what @swaroopsalvi is talking about and it and it doesn't say placeholder for creating a new one, but not sure if it is in scope

for 2) I'm not sure if we are comparing the create semantic vs edit semantic screens with the stamp area @swaroopsalvi.  We might need a meeting to go over that because I don't follow what you are comparing.  When I run her branch locally the create shows an empty stamp and that is what it should do.

dholubek (approved):
looks good

PR #339, opened 2025-03-24
IIA-1590 - Clear function implementation for Next gen search pane

bhharsh13:
Implemented the Clear function for Next gen search pane.
https://ikmdev.atlassian.net/browse/IIA-1590
Clear searchText Field
After clear

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.49.0-SNAPSHOT</version>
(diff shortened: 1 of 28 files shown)


dukke:
@bhharsh13 the text inside the textfield needs to be cleared only when you press the clear (x) button

dukke:
> isn't the cursor supposed to change on the text vs the 'X' button? It is an arrow in both places... @bhharsh13 @dukke

Good point! That is correct. It should be a "text editing" cursor when hover the textfield and the "default" cursor when over the x button.

dukke (approved):
Looks good to me! We can create another ticket for the more complex task of creating a textfield with a more "proper" clear feature.

PR #344, opened 2025-03-26
Feature/finished/iia 1130 default values

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1130

image

image

also added a minor fix to the Integer component to allow a zero.  It is filtering out leading zeros, so "01" is not allowed, "1" is allowed, but it was being too aggressive and not allowing just "0".

this is missing the reference component as you can see.


bhharsh13 (approved):
Looks good to me

PR #348, opened 2025-03-27
Feature/iia 1579 semantic title

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1579

image


swaroopsalvi (approved):
Looks good approved.

carldea (approved):
Looks good to me. I will approve and merge.

We should keep an eye to see if the Figma designs intended the Purpose's text width wrapping is an issue. I think the figma designs is shown with a small width.

PR #350, opened 2025-03-31
Add Concept Navigator control to Komet

jperedadnr:
This PR adds the KLConceptNavigatorControl (that is being developed here: https://github.com/jperedadnr/komet/tree/concept-navigator and can be tested only from the SamplerApp) into the kView module, so it can be run within Komet.

So far, the control can be reached from the Fav/Heart icon:

image

The integration is more or less complete, however dragging and dropping a selection of concepts is not working yet, as there is only one single KLDropRegion (and not more like in the Figma design).


carldea (approved):
@jperedadnr  Amazing work! This work is practically pixel perfect. I like the logic of state management on the various pseudo states.

PR #352, opened 2025-04-01
Feature/finished/iia 1706 semantic full title

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1706

max value and hgrow to allow the text to expand when the window expands

image


swaroopsalvi (approved):
Looks good.

PR #355, opened 2025-04-02
Feature/finished/iia 1130 rc anon

dholubek:
This is a follow up PR for https://ikmdev.atlassian.net/browse/IIA-1130

We are using anonymous concept.  This will likely affect the JavaFX UI components

image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.49.0-IIA-1130-rc-anon-SNAPSHOT</version>
+        <version>1.49.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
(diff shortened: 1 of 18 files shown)


dukke:
Given yesterdays talk with Keith and our talk with Ryan at the architecture call we'll likely need to discuss this further to reach a consensus on what our approach will be.
I think our understanding of autosave and saving non valid (empty) fields has changed. It seems our approach will not allow empty fields or reference components to be autosaved and in that case we might not need Anonymous Concept anymore (?)

dukke (commented):
Given yesterdays talk with Keith and our talk with Ryan at the architecture call we'll likely need to discuss this further to reach a consensus on what our approach will be.
I think our understanding of autosave and saving non valid (empty) fields has changed. 
There are a few approaches that Keith mentioned we can take...we'll need to decide on one.

carldea (approved):
Looks good to me! thank you.

carldea:
> Given yesterdays talk with Keith and our talk with Ryan at the architecture call we'll likely need to discuss this further to reach a consensus on what our approach will be. I think our understanding of autosave and saving non valid (empty) fields has changed. There are a few approaches that Keith mentioned we can take...we'll need to decide on one.

I think it was decided to use the Anonymous concept for now. I hope this doesn't cause more confusion. Let's just get the happy path first (initial create).

PR #356, opened 2025-04-03
Nexus

jlgrock:
Adding a number of docker-compose targets including nexus, gitea, and a homepage that can be accessed via http://localhost when doing a `docker compose up`


docker/Dockerfile-komet-web
```diff
@@ -0,0 +1,29 @@
...
+
+COPY .. /komet/
+
+WORKDIR /komet/
+
+RUN ls /komet/
+RUN ./mvnw clean install -Dmaven.javadoc.skip=true -Dmaven.test.skip=true
+RUN ./mvnw clean -f application -Pjpro jpro:release
+
+FROM azul/zulu-openjdk-alpine:23-latest
```
carldea, line 11: What's the difference between line 11 and line 1 (besides a different layer)?

jlgrock: It runs an additional profile after the installation on the last module.  It might be combined but JPro guys ran into issues when I combined, and they changed it back to get it publishing.

besidev: Since **Komet** project uses the Maven wrapper for builds, we recommend the following Docker image changes:

1. Line 1: Use `azul/zulu-openjdk-alpine:23-latest` as the builder image
2. Line 11: Use `azul/zulu-openjdk-alpine:23-jre-headless-latest` for the runtime image

This ensures consistency with the same Java version and vendor across both stages, while optimizing the runtime container by including only the JRE headless components needed for execution.

While we have already tested these changes successfully on our side, please perform the required testing on your side as well.

jlgrock: Sorry, he is correct.  You don't want a JDK for runtime.  Just for building.  I misunderstood the question.

carldea (approved):
Great work!
Looks good to me. 
I will approve and merge as this doesn't effect the installed application.

**Note:**
In the future we should examine ways to use jlink and the dockSlim tool to make thinner JRE runtimes and docker images respectively. Based on our profiling the app while using snomedct it is very memory heavy. Reducing the runtime footprint would be ideal.

docker/Dockerfile-komet-web
carldea: Is there more work to do or can I merge this PR? I didn't want to hold things up.

PR #357, opened 2025-04-03
changes to show an overlay to see JVM resources

jlgrock:
Creating an overlay that can be accessed via the view menu to see current resource usage.  Currently shows memory and CPU.

### DIFF
--- a/application/src/main/java/dev/ikm/komet/app/App.java
+++ b/application/src/main/java/dev/ikm/komet/app/App.java
@@ -40,6 +40,7 @@ import static dev.ikm.komet.preferences.JournalWindowSettings.JOURNAL_TITLE;
 import static dev.ikm.komet.preferences.JournalWindowSettings.JOURNAL_WIDTH;
 import static dev.ikm.komet.preferences.JournalWindowSettings.JOURNAL_XPOS;
 import static dev.ikm.komet.preferences.JournalWindowSettings.JOURNAL_YPOS;
+
 import de.jangassen.MenuToolkit;
 import de.jangassen.model.AppearanceMode;
 import dev.ikm.komet.details.DetailsNodeFactory;
@@ -85,6 +86,8 @@ import dev.ikm.tinkar.common.alert.AlertStreams;
 import dev.ikm.tinkar.common.binary.Encodable;
 import dev.ikm.tinkar.common.service.PrimitiveData;
 import dev.ikm.tinkar.common.service.TinkExecutor;
+import javafx.animation.KeyFrame;
+import javafx.animation.Timeline;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.beans.property.SimpleObjectProperty;
@@ -109,8 +112,10 @@ import javafx.scene.layout.StackPane;
 import javafx.scene.layout.VBox;
 import javafx.scene.paint.Color;
 import javafx.scene.text.Font;
+import javafx.stage.Modality;
 import javafx.stage.Stage;
 import javafx.stage.StageStyle;
+import javafx.util.Duration;
 import org.carlfx.cognitive.loader.Config;
 import org.carlfx.cognitive.loader.FXMLMvvmLoader;
 import org.carlfx.cognitive.loader.JFXNode;
@@ -122,6 +127,10 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
+import java.lang.management.ManagementFactory;
+
+import com.sun.management.OperatingSystemMXBean;
+
 import java.time.LocalDateTime;
 import java.time.Year;
 import java.time.ZoneId;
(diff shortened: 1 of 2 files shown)


application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -901,6 +921,78 @@ private MenuItem createClassicKometMenuItem() {
...
+        });
+        resourceUsageItem.setAccelerator(classicKometKeyCombo);
+        return resourceUsageItem;
+    }
+
+    /**
+     * Show the resource usage overlay.
+     */
+    private void showResourceUsageOverlay() {
+        overlayStage = new Stage();
```
carldea, line 947: Can we a close strategy?
**Reason:** memory leaks if the menu option gets called again a new Stage window is created with an running animation.

**Suggestion**
You may want to check if overlayStage was already instantiated and hide or add the call back `setOnCloseRequest()` to clean up (stop animation timer).

application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -901,6 +921,78 @@ private MenuItem createClassicKometMenuItem() {
...
+        Scene overlayScene = new Scene(overlayContent, 300, 200, Color.TRANSPARENT);
+        overlayStage.setScene(overlayScene);
+        overlayStage.show();
+
+        // Create and start the timeline to update resource usage
+        resourceUsageTimeline = new Timeline(new KeyFrame(Duration.seconds(1), event -> {
+            updateResourceUsage(cpuUsageLabel, memoryUsageLabel);
+        }));
+        resourceUsageTimeline.setCycleCount(Timeline.INDEFINITE);
+        resourceUsageTimeline.play();
```
carldea, line 971: Can we add code to clean up when the user clicks on the close button?
JavaFX's `hide() and close()` are essentially the same and you can not reshow a `Stage` after it has been closed.
**Reason:** memory leaks if the menu option gets called again a new Stage window is created with an running animation. hide and close aren't the same. In JavaFX once you close a Stage you can not show again. if the the code is called again a new Stage will be created, but the previous stage's animation is still polling the JMX info.

**Suggestion**
`overlayStage.setOnCloseRequest(...)` to clean up referenced objects (stop animation `Timeline` timer).

carldea (commented):
@jlgrock Nice work! I like the JMX info.
I added some feedback.

application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -901,6 +921,78 @@ private MenuItem createClassicKometMenuItem() {
...
+     * Update the resource usage labels with CPU and memory usage.
+     *
+     * @param cpuUsageLabel the label to be used for the CPU usage
+     * @param memoryUsageLabel the label to be used for the memory usage
+     */
+    private void updateResourceUsage(final Label cpuUsageLabel, final Label memoryUsageLabel) {
+        java.lang.management.OperatingSystemMXBean osBean = ManagementFactory.getOperatingSystemMXBean();
+        double cpuLoad = -1;
+        if (osBean instanceof OperatingSystemMXBean) {
+            cpuLoad = ((OperatingSystemMXBean) osBean).getCpuLoad() * 100;
```
carldea, line 984: Can we use Java's new pattern matching?
```java
if (osBean instanceof OperatingSystemMXBean osMXBean) {
   cpu = osMXBean.getCpuLoad() * 100;
   ...
```

carldea: Actually I don't think there is a need to check if it is an instance of `java.lang.management.OperatingSystemMXBean` since the object is already of that type.

application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -901,6 +921,78 @@ private MenuItem createClassicKometMenuItem() {
...
+
+            long totalMemory = Runtime.getRuntime().totalMemory();
+            long freeMemory = Runtime.getRuntime().freeMemory();
+            long usedMemory = totalMemory - freeMemory;
+
+            cpuUsageLabel.setText(String.format("CPU Usage: %.2f%%", cpuLoad));
+            memoryUsageLabel.setText(String.format("Memory Usage: %d MB / %d MB",
+                    usedMemory / (1024 * 1024), totalMemory / (1024 * 1024)));
```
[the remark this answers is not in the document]
jlgrock: sure. will add

application/src/main/java/dev/ikm/komet/app/App.java
jlgrock: Adding now!

application/src/main/java/dev/ikm/komet/app/App.java
jlgrock: It's confusing, but I actually do.  The one class is from the sun package and the other is from the java package, even though they have the same name.  The sun one is a subclass of the java one and has additional functionality to get the CPU, which isn't available on all systems, so it can't be used on IoT and things like that.  Thus, the java one won't let me get the CPU load, unfortunately, so I have to down convert to get it, and it's always worth a check to see if it is the correct class when down casting.

I hadn't used the enhanced instanceof though, so I incorporated that.  That's pretty nice.

jlgrock:
updates made!

PR #358, opened 2025-04-03
Feature/finished/iia 1720 remove component exception fix

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1720

I understand that coding to exception logic is not ideal, but the toString() queries the DB and throws it.

IIA-1720 remove component


dholubek:
@swaroopsalvi 's fix is the correct one and will close this PR

PR #360, opened 2025-04-04
Feature/finished/iia 1729 disable submit button on blank fields

swaroopsalvi:
Editing : Submit button enabled.
image

One or more blank fields: Submit button is disabled
image


kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
```diff
@@ -170,7 +170,7 @@ public final void setOnAddConceptAction(EventHandler<ActionEvent> value) {
      * A property with an action to be executed when the user clicks on the remove button. By default, it will
      * remove the entity that was added.
      */
-    private final ObjectProperty<EventHandler<ActionEvent>> onRemoveActionProperty = new SimpleObjectProperty<>(this, "onRemoveAction", e -> setEntity(EntityProxy.make(EMPTY_NID)));
+    private final ObjectProperty<EventHandler<ActionEvent>> onRemoveActionProperty = new SimpleObjectProperty<>(this, "onRemoveAction", e -> setEntity(null));
```
swaroopsalvi, line 173: When we remove the entity object from the editable pane we need to set it to blank.

dholubek (approved):
looks good

PR #362, opened 2025-04-07
Feature/finished/iia 1728 c list add entry

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1728


carldea (approved):
Looks good to me. I think from a UX standpoint, when the search moves to the top, the user will not be surprised (works the same way).

dukke:
> Looks good to me. I think from a UX standpoint, when the search moves to the top, the user will not be surprised (works the same way).

Yes!  That "add button" is going to disappear in the next redesign :)
Thanks!

PR #364, opened 2025-04-08
Feature/finished/iia 1397 enable disable submit

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1397

This was already mostly working except for empty collections

image

image

image


dholubek:
we can save empty collections... this PR is invalid so I'm going to close it

PR #365, opened 2025-04-08
Feature/finished/iia 1719 overlapping context menus

jdsmithsos:
Before:

image

After fix:

image


bhharsh13 (commented):
Looks good to me.
If we can add the link to story/defect in the PR, it would be convenient to navigate.

bhharsh13 (approved):

PR #366, opened 2025-04-08
Updated ubuntu to 24.04 for the installer builds

OmOmofonmwan:


dukke:
Looks good to me!
I was pointing out this error as builds started to give errors because the Ubuntu version was outdated (they gave an error when generating installer for Ubuntu). This change seems to address that.

Thanks for quickly fixing this @OmOmofonmwan

carldea (approved):
Looks good to me. Thanks @OmOmofonmwan !

PR #368, opened 2025-04-09
Feature/finished/iia 1597 wrong number of pattern fields

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1597

creating a pattern, adding the first field:
image

adding the second field:
image

click the globe icon (fields stay at 2):
image


jdsmithsos (approved):
looks good to me

carldea (approved):
Looks good to me. 
Code looks relatively the same as before. It appears in certain situations it's off by one.

dukke (commented):
Nice work!
Just had some comments that are not hyper important, but I would say should be good changes.

PR #371, opened 2025-04-10
Feature/finished/iia 883 type ahead search control

dukke:
fixes https://ikmdev.atlassian.net/browse/IIA-883

## Video showing it running in the sampler
https://github.com/user-attachments/assets/70839fbf-006b-4cc9-ae3d-167099450988

Notes: When I press the enter key the right textfield in the sample is updated with the search result

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.50.0-IIA-883-TypeAheadSearch-Control-SNAPSHOT</version>
+        <version>1.50.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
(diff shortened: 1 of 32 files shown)


dukke:
@carldea update: added a property to define the debounce time instead of it being defined in a static field.

jdsmithsos:
I'm new here so I don't know the requirement that required a custom autocomplete text field to be created.  ControlsFX has an autocomplete text field that I have used.  It must not have met the requirements for Komet, and could not be extended for the accessing the data source.

dukke:
> I'm new here so I don't know the requirement that required a custom autocomplete text field to be created. ControlsFX has an autocomplete text field that I have used. It must not have met the requirements for Komet, and could not be extended for the accessing the data source.

Good observation but indeed ControlsFX popup for auto complete doesn't meet our requirements. As a side note this control is also a full fledged textfield instead of a popup that attaches to a control (like in ControlsFX).
Thanks

dholubek (approved):
looks good

carldea:
Looks good! Thank you!

PR #372, opened 2025-04-11
Feature/finished/iia 1640 add error message to float editable field

jdsmithsos:
Summary of changes:

- Added the calls to setShowError()
- Modified the layoutChildren() to set the width of the error label to match the width of the text field, which is what the Integer control skin does
- Modified KLIntegerControlSkin to make the ERROR_PSEUDO_CLASS and ERROR_DURATION to be public so the KLFloatControlSkin class can use
- Modified float css to match integer css

Jira ticket:

https://ikmdev.atlassian.net/browse/IIA-1640

Before fix:

image

With fix:

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLFloatControlSkin.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLFloatControlSkin.java
@@ -2,13 +2,11 @@ package dev.ikm.komet.kview.controls.skin;
 
 import dev.ikm.komet.kview.controls.KLFloatControl;
 import javafx.animation.PauseTransition;
-import javafx.css.PseudoClass;
 import javafx.geometry.Insets;
 import javafx.scene.control.Label;
 import javafx.scene.control.SkinBase;
 import javafx.scene.control.TextField;
 import javafx.scene.control.TextFormatter;
-import javafx.util.Duration;
 import javafx.util.Subscription;
 
 import java.text.MessageFormat;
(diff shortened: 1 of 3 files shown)


kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLFloatControlSkin.java
```diff
@@ -79,10 +77,12 @@ public KLFloatControlSkin(KLFloatControl control) {
                     // discard if we have a valid value, but it is infinite or NaN
                     if (Double.isInfinite(value) || Double.isNaN(value)) {
                         errorLabel.setText(MessageFormat.format(resources.getString("error.float.text"), newText));
+                        control.setShowError(true);
                         return null;
                     }
                 } catch (NumberFormatException nfe) {
-                    // ignore
+                    errorLabel.setText(MessageFormat.format(resources.getString("error.float.text"), newText));
+                    control.setShowError(true);
```
dukke, line 85: Is there some way we can do this logic without using exception handling? Exceptions can be expensive and usually I think should be used for exceptional error cases (not as part of the program's "normal logic").

dukke (commented):
Good work!! 👍
Just had one comment...

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLFloatControlSkin.java
jdsmithsos: This was the way it was coded.  The KLIntegerControlSkin does the same thing with a NumberFormatException for integer values.

dukke (approved):
Looks good. We can perhaps then create a ticket to do John's suggestion

bhharsh13 (approved):
Looks good to me

PR #375, opened 2025-04-14
Feature/finished/iia 1808 exception on semantic submit

dholubek:
The NPE was because in create mode the observableFields and their associated editActions aren't set up yet, so when we submit we want to update the left and we need them to be ready to receive the updates.

add a new semantic, notice only 3 semantics are there
adding new semantic

user sees the default values
default values

user then enters values
user entered values

user clicks submit to create the semantic and the values go to the left
submit moves values to the left

we reopen the journal and the new semantic is there (refresh of the patterns is IIA-1785)
reopen journal and a new semantic is there

same for component, also saves
component saves

still need to capture the reference component but that effort can be done in IIA-1396

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.50.0-IIA-1808-exception-on-semantic-submit-SNAPSHOT</version>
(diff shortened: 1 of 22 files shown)


bhharsh13 (approved):
Looks good to me

carldea (approved):
Looks good to me.
I do like the refactor of the setup code as another private method.

PR #376, opened 2025-04-14
Feature/finished/iia 1785 semantic save refresh pattern nav

dholubek:
Fire event to reload the pattern window.  It doesn't preserve what is expanded but that was never in the AC

image


carldea (approved):
Looks good to me!
Thanks.

PR #377, opened 2025-04-14
IIA-1578 Disable the  edit the Reference Component of an existing semantic

bhharsh13:
Implemented the fix for disabling the edit the Reference Component of an existing semantic once that semantic has been submitted. PR is under review.

Validated implementation on local and changes looks good. 
PFA for reference.
Disable Reference Component Icon
Enable Button for new Semantic

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
@@ -63,9 +63,11 @@ import dev.ikm.tinkar.coordinate.language.calculator.LanguageCalculator;
 import dev.ikm.tinkar.coordinate.stamp.calculator.Latest;
 import dev.ikm.tinkar.coordinate.stamp.calculator.StampCalculator;
 import dev.ikm.tinkar.entity.ConceptEntity;
+import dev.ikm.tinkar.entity.EntityService;
 import dev.ikm.tinkar.entity.EntityVersion;
 import dev.ikm.tinkar.entity.PatternEntityVersion;
 import dev.ikm.tinkar.entity.PatternVersionRecord;
+import dev.ikm.tinkar.entity.SemanticEntity;
 import dev.ikm.tinkar.entity.SemanticEntityVersion;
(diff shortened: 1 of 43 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
```diff
@@ -379,6 +379,13 @@ private void setupReferenceComponentUI(Latest<SemanticEntityVersion> semanticEnt
         ObjectProperty<EntityFacade> refComponentProp = genEditingViewModel.getProperty(REF_COMPONENT);
         EntityFacade refComponent = refComponentProp.get();
 
+        //Disable the  edit the Reference Component of an existing semantic once submitted
+        if(refComponent != null){
+            addReferenceButton.setDisable(true);
+        }else{
+            addReferenceButton.setDisable(false);
+        }
```
dholubek, line 387: can we have spaces outside of the if parens and the else?

```
        if (refComponent != null) {
            addReferenceButton.setDisable(true);
        } else {
            addReferenceButton.setDisable(false);
        }
```

dholubek (changes requested):
had a code styling comment

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
bhharsh13: Added the spaces outside of if and else. Thanks Doug.

dukke: Can we collapse those lines into this: 
addReferenceButton.setDisable(refComponent != null);

dukke (approved):
Looks good to me! 👍

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
bhharsh13: modified the code

dholubek (approved):
looks good

PR #378, opened 2025-04-14
Docker port variability

jlgrock:


carldea (approved):
Looks good to me.
isn't port 22 reserved? 
BTW I'm approving and merging.

jlgrock:
Yes and no.  Most of this is internal networking.  The port 22 will be exposed eventually, but we need TLS cert generation first

PR #379, opened 2025-04-14
IIA-1796 IIA-1816 fixed both bugs by using the Window rather than the…

jdsmithsos:
… Node to attach the ContextMenu to, which allows auto close when the mouse is pressed outside of the contextMenu

Summary of changes:  changed using the panel Node as the parent of the contextMenu to using the Window of the Node

https://ikmdev.atlassian.net/browse/IIA-1796
https://ikmdev.atlassian.net/browse/IIA-1816

IIA-1796 before:

image

IIA-1816 before:

image

IIA-1796 - using Window rather than Node, with right mouse click:

image

IIA-1796 - pressing the mouse on the parent Node, the contextMenu auto hides:

image

IIA-1816 - using Window rather than Node, with right mouse click on a Field:

image

IIA-1816 - pressing the right mouse button again on the same Field, the original context menu is hidden, and a new context menu is created, without multiple context menus being displayed:

image


bhharsh13 (approved):
Looks good to me

carldea (approved):
Looks good to me!

PR #381, opened 2025-04-15
IIA-1822 round two of IIA-1640, where the text was a bug because…

jdsmithsos:
…it wasn't consistent with the Integer error text

Jira bug:  https://ikmdev.atlassian.net/browse/IIA-1822

Summary of changes:

- Changed the error text per bug's description
- Removed the unnecessary formatter because the error text does not have a parameter
- Added another show error at the NumberFormatException, which was missed in the original defect

Before fix:

image

After fix with consistent error message:

image


carldea (approved):
looks good to me.

PR #384, opened 2025-04-16
IIA-1831 changed to match Integer where if the input string value is …

jdsmithsos:
…empty to not show the error label

Summary of changes:  re-arranged the first if statement to be coded like the KLIntegerControlSkin where newText.isEmpty() and !newText.isEmpty() is checked

Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-1831

Before fix:

image

After fix:

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLFloatControlSkin.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLFloatControlSkin.java
@@ -66,22 +66,22 @@ public class KLFloatControlSkin extends SkinBase<KLFloatControl> {
             // - adding '-' or '+' if empty or ends in e/E
             // - back when text ends in [-/+, e, e-/+, E, E-/+]
             // - pattern
-            if (newText.isEmpty() ||
-                    (hasExponent(newText) && !hasExponent(oldText) && !oldText.isEmpty()) ||
-                    (("-".equals(addedText) || "+".equals(addedText)) && (oldText.isEmpty() || endsWithExponent(oldText))) ||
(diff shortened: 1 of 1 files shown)


bhharsh13 (approved):
Looks good to me

jdsmithsos:
> @jdsmithsos why is the after screen shot showing no error if the user blanks out the Float field? Shouldn't it message the user that it is required?

Carl, this was the bug that was reported and fixed.  The Float field did not behave like the Integer field, where if the value is blank, there is no error message.
What you are saying is that for all required fields (which are all because null column values aren't allowed?) that there should be an error displayed when the text field is empty?
That would be a different bug...  If written should address all text text fields.

carldea:
> itten should address all text fields.

You are probably correct based on the Integer field. but I think it should complain (both Integer and Float), and tell the user the field is required. All fields must have valid (populated) values when validating in support of enabling or disabling the submit button. In the case of String empty string is valid.

If a user, had nothing entered into Float, Integer, Or BigDecimal, should the submit button allow the user to hit submit? I don't think it should. But also provide feedback so the user would correct the UI form.

jdsmithsos:
> > itten should address all text fields.
> 
> You are probably correct based on the Integer field. but I think it should complain (both Integer and Float), and tell the user the field is required. All fields must have valid (populated) values when validating in support of enabling or disabling the submit button. In the case of String empty string is valid.
> 
> If a user, had nothing entered into Float, Integer, Or BigDecimal, should the submit button allow the user to hit submit? I don't think it should. But also provide feedback so the user would correct the UI form.

I talked to Doug about next steps on this, and because the code fixes the original bug reported, this PR will be merged, and a new Bug created to address the blank text fields error, because all fields are required, meaning no blank fields allowed.

PR #385, opened 2025-04-16
IIA-1536 Lift Item from Component List and Component Set When Reordering

bhharsh13:
Fixed the implementation -
When I drag the concept away from its original position, I should see the concept be “lifted” from its original position so that the original position is empty.
Component List Drag fix

Component Set Drag

Working on implementing -
When I drag the concept back to its original position, I should see the original position focused with a green-dotted outline so that I know I am selecting that position again for the concept.


carldea (approved):
Excellent work @bhharsh13!

carldea (changes requested):
Let's get more feedback from @dukke before we approve. Was there additional fix related to pseudo states?

dukke:
> Let's get more feedback from @dukke before we approve. Was there additional fix related to pseudo states?

@carldea I worked with Bhavani on this first part of the ticket - getting the component to appear as if it "lifts", leaving an empty space where it was. 
According to the AC they also want a white background box with a border to appear when the user is hovering over a component position (still to be implemented).

carldea:
> @carldea I worked with Bhavani on this first part of the ticket - getting the component to appear as if it "lifts", leaving an empty space where it was.
According to the AC they also want a white background box with a border to appear when the user is hovering over a component position (still to be implemented).

@dukke @bhharsh13 could you put the PR in draft since you are still working on it. :-)

dukke:
@bhharsh13 @carldea Closing this PR as another PR that added this feature as already been merged.

PR #386, opened 2025-04-17
Restrict mouse right-click and increase font size

jperedadnr:
Fixes some issues in the concept navigator (treeView and search controls):
- restrict right-click events
- increase font size


carldea (approved):
Looks good!

PR #389, opened 2025-04-19
Feature/finished/kec changeset

kec:
Made changes necessary to include change set provider in the application. Added it to the module path of the application, and added the dependency. I know that it will fail on failing to find the version of tinker-core that I specified. Ideally, you'll need to get a release of thinker-core that include my other pull request made today, and then updated the version to that one.


carldea (approved):
This looks good to me!

PR #400, opened 2025-04-23
IIA-1871 Component List/ Set drag position fix

bhharsh13:
Fixed the implementation -
When I drag the concept away from its original position, I should see the concept be “lifted” from its original position so that the original position is empty.
https://ikmdev.atlassian.net/browse/IIA-1871

Component List Drag fix
Component Set Drag fix

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
@@ -247,7 +247,6 @@ public class KLComponentListControlSkin<T extends IntIdCollection> extends SkinB
 
         componentControl.addEventHandler(DragEvent.DRAG_DONE, _ -> {
             componentControl.setVisible(true);
-            componentControl.removeEventHandler(MouseEvent.MOUSE_MOVED, onMouseMoved);
         });
 
         Label numberLabel = createNumberLabel(componentControl);
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
@@ -238,14 +238,11 @@ public class KLComponentListControlSkin<T extends IntIdCollection> extends SkinB
         EventHandler<MouseEvent> onMouseMoved = mouseEvent -> {
             if (mouseEvent.getTarget() == componentControl) {
                 componentControl.setVisible(true);
-            } else {
-                componentControl.setVisible(false);
             }
         };
 
         componentControl.addEventHandler(MouseEvent.DRAG_DETECTED, _ -> {
             componentControl.setVisible(false);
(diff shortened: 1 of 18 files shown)


kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
```diff
@@ -233,6 +235,24 @@ private void createComponentUI(int nid) {
 
         componentControl.showDragHandleProperty().bind(componentControl.hoverProperty());
 
+        EventHandler<MouseEvent> onMouseMoved = mouseEvent -> {
+            if (mouseEvent.getTarget() == componentControl) {
+                componentControl.setVisible(true);
+            } else {
+                componentControl.setVisible(false);
+            }
+        };
```
dukke, line 244: we can remove this code above. It's not needed to make the component invisible (empty spot) when the user is dragging.

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
```diff
@@ -233,6 +235,24 @@ private void createComponentUI(int nid) {
...
+
+        componentControl.addEventHandler(MouseEvent.DRAG_DETECTED, _ -> {
+            componentControl.setVisible(false);
+            componentControl.addEventHandler(MouseEvent.MOUSE_MOVED, onMouseMoved);
```
dukke, line 248: we can remove this line of code that add the event handler and only keep the one that sets the visibility

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
```diff
@@ -233,6 +235,24 @@ private void createComponentUI(int nid) {
...
+        });
+
+        componentControl.addEventHandler(DragEvent.DRAG_DONE, _ -> {
+            componentControl.setVisible(true);
+            componentControl.removeEventHandler(MouseEvent.MOUSE_MOVED, onMouseMoved);
```
dukke, line 253: we can remove this line that removes the event handler

dukke (changes requested):
Made a few suggestions.. 
So, to my understanding (and as a note to everyone who reviews this) we will just take care of making the component invisible (leaving an empty spot) while the user is drag and dropping for now.
Thanks!

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
bhharsh13: removed the addEventHandler

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentListControlSkin.java
bhharsh13: removed the else block

dukke (approved):
Looks good! Thanks!

PR #401, opened 2025-04-23
Feature/finished/iia 1806 submit default on create

dholubek:
created a carve out in the logic so that when you are in create mode and the default values are exactly what they are then you are allowed to just save them as is.

image


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
```diff
@@ -108,9 +109,16 @@ public class SemanticFieldsController {
     private void enableDisableSubmitButton(){
         //Disable submit button if any of the fields are blank.
         boolean disabled = checkForEmptyFields();
-        if(!disabled){
+        if (!disabled) {
             int uncommittedHash = calculteHashValue(observableFields);
-            disabled = (committedHash == uncommittedHash);
+            // carve out in the logic so that when you are in create mode and the default values
+            // are exactly what they are then you are allowed to just save them as is
+            if ((genEditingViewModel.getPropertyValue(MODE) == CREATE) &&
```
swaroopsalvi, line 116: Can we remove `if and else` condition and always use view model for both edit and create mode.
Also not sure which one is better using  `== ` or `.equals` since we are comparing integer ?

So something like:

`disabled = uncommitted == genEditingViewModel.getPropertyValue(DEFAULT_FIELDS_HASH))`

swaroopsalvi (commented):
Liked the MVVM approach. Just have one comment.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
dholubek: the reason I didn't do that was because in EDIT mode, if they don't change any values then the submit should be disabled... so if it just equals the default hash then it would be enabled in EDIT mode when it should be disabled.

swaroopsalvi (approved):
Thank you Doug

PR #402, opened 2025-04-23
Feature/finished/finished/view calculator invalidate cache impl

bhharsh13:
We were able to fix the issue related to-  [[[IIA-1599] Import dataset - User edits the concept data and exported dataset. After importing the dataset, the edited Fully Qualified Name (FQN) and Other name for a concept are not reflected. - Jira](https://ikmdev.atlassian.net/browse/IIA-1599) ](https://ikmdev.atlassian.net/browse/IIA-1599) . We put the thread to sleep for 5 milliseconds and then  invoke CachingService.clearAll() for refreshing GLOBAL CACHE. I have commited the changes in the branch and sharing the PR. Dea, Carl needs to review the PR if further refinement is needed. I validated the changes on my local and looks good.

FQN OtherName update post import

[IIA-1599]: https://ikmdev.atlassian.net/browse/IIA-1599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


artifact/src/main/java/dev/ikm/komet/artifact/ArtifactImportController.java
```diff
@@ -108,7 +113,15 @@ void initiateProtobufTransform(ActionEvent event) {
...
         importProgressBar.progressProperty().unbind();
         importProgressBar.progressProperty().bind(importTask.progressProperty());
-        ProgressHelper.progress(importTask, "Cancel Import");
+
+        Future future =  ProgressHelper.progress(importTask, "Cancel Import");
+        try {
+            Thread.sleep(5000);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
```
carldea, line 122: Can we remove the sleep code in favor of a proxy pattern (overridden method)? Let me explain:

In Java any non final methods can be overridden during creation. Such as the following:
```java
Object x = new Object() {
   public String toString() {
       return super.toString() + " hello";
   }
};

 x.toString(); 
```
outputs:
```
1@4883b407 hello
```
In line 111 reads: 
```java
LoadEntitiesFromProtobufFile loadEntities = new LoadEntitiesFromProtobufFile(selectedFile);
```
Override the `compute()` method like the following:
```java
LoadEntitiesFromProtobufFile loadEntities = new LoadEntitiesFromProtobufFile(selectedFile) {
   public EntityCountSummary compute() {
       EntityCountSummary summary = super.compute();
       EvtBusFactory.getDefaultEvtBus().publish(CALCULATOR_CACHE_TOPIC, new RefreshCalculatorCacheEvent(event.source(), GLOBAL_REFRESH));
       return summary;
   }
};
```
Above code will proxy or intercept the call to compute to do the following:

1. Call the `LoadEntitiesFromProtobufFile.compute()` and assign to a variable.
2. When finished (complete the import process) publish event to refresh the cache.
3. Return the summary to the caller.

The goal was to "know" when **import** was finished then emit the event. 

### Optional steps (Create a ticket) this might be out of scope of this defect.
In a later ticket we could enhance an already summoned concept window to refresh latest **fqn** or **other names**. 

e.g., If there are new imports a subscriber (listener) could check a concept if any semantics that have a newer date (new versions) if so refresh the ui, and if not ignore refreshing ui.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
```diff
@@ -408,6 +414,15 @@ public void initialize() {
         });
 
         workspace.addEventHandler(MouseEvent.MOUSE_CLICKED, this::onMouseClickedOnDesktopSurfacePane);
+        refreshCalculatorEventSubscriber = evt -> {
+            LOG.info("Refresh Calculator Event");
+            if(evt.getEventType() == GLOBAL_REFRESH) {
+                LOG.info("Global Refresh EventYpe...");
```
carldea, line 420: update the log message to output the `GLOBAL_REFRESH`. tip: Debugging event messages it's often a good idea to show the name of the event types and topics in the console logs.

carldea (approved):
Looks good to me!

Later on we should revisit the code to move validation logic into the view model.

PR #404, opened 2025-04-24
IIA-1528 using a common editable-title-label rule in kview.css for al…

jdsmithsos:
…l editable labelFields, to have consistent horizontal spacing between the title and editable field

https://ikmdev.atlassian.net/browse/IIA-1528

Summary of changes:

- Common css rule for editable field titleLabel is now in kview.css, named .editable-title-label
- Set the bottom padding to 10px, which must be at least the same size as the SPACE_BETWEEN_COMPONENTS constant, which is set to 10
- The following editable KL?ControlSkin classes were updated to use the new .editable-title-label css rule, via their corresponding css files:
- KLBooleanControlSkin
- KLByteArrayControlSkin (not currently used, but change has been applied)
- KLComponentControlSkin
- KLComponentListControlSkin
- KLFloatControlSkin
- KLImageControlSkin
- KLIntegerControlSkin
- KLStringControlSkin

Before without use of css:

image

image

image

The new css rule:  (with green background just to show the change clearly)

image

The changes to the titleLabel in the KL?ControlSkin classes:

image

image

image


dholubek (approved):
looks good

carldea (approved):
Looks good to me!

swaroopsalvi (approved):
Awesome

PR #405, opened 2025-04-24
Feature/finished/iia 1738 emergency pr classic komet f ix

swaroopsalvi:


carldea (approved):
Excellent! Good eye @swaroopsalvi 
Glad classic Komet now works like before the new autosave feature.

PR #408, opened 2025-04-28
Feature/finished/iia 1887 full reasoner only

dholubek:
IIA-1887, turn off incremental reasoner

image


carldea (approved):
Looks good to me. Makes sense.

PR #409, opened 2025-04-28
Feature/finished/iia 1873 refresh component on left from edit submit

dholubek:
IIA-1873, on edit and component change, then submit, the concept wasn't getting refreshed on the left.  now it does

image

image


carldea (approved):
Looks good to me. This should also work for the reference component section.
Thanks!

PR #410, opened 2025-04-29
feature/finished/IIA-1855-exception-on-create-test-pattern-3

jdsmithsos:
Jira ticket:

https://ikmdev.atlassian.net/browse/IIA-1855

Before, when creating a Semantic from the Test Pattern 3 Pattern, a NullPointerException was occurring in KlFieldHelper.addReadOnlyBlankControlsToContainer() because the TinkarTerm.BYTE_ARRAY_FIELD was not supported for an Image.

image

After adding a check for the TinkarTerm.BYTE_ARRAY_FIELD for an image, the semantic is successfully created without exception:

image


dholubek (approved):
looks good

PR #411, opened 2025-04-29
Feature/finished/iia 1820 supported pattern datatypes

dholubek:
This restricts the options of the pattern creation field's datatypes to ones that are supported.  It works with the new dataset (and the old dataset because that treats an image like a byte array)

https://repo1.maven.org/maven2/dev/ikm/data/tinkar/tinkar-example-data/1.2.0+1.1.5/tinkar-example-data-1.2.0+1.1.5-reasoned-sa.zip

shows up in drop down and is handled when you create it
image field in pattern dropdown

with existing test pattern 3
image field with new dataset

works with the old dataset
image field with old dataset


kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -224,10 +234,10 @@ public static List<KLReadOnlyBaseControl> addReadOnlyBlankControlsToContainer(Pa
...
                 control = new KLReadOnlyComponentSetControl();
+            } else if (fieldDefinitionRecord.dataTypeNid() == TinkarTerm.IMAGE_FIELD.nid()) {
+                control = new KLReadOnlyImageControl();
             } else if (fieldDefinitionRecord.dataTypeNid() == TinkarTerm.BYTE_ARRAY_FIELD.nid()) {
-                //TODO: We're using BYTE_ARRAY for the moment for Image data type
-                //TODO: using IMAGE_FIELD would require more comprehensive changes to our schema (back end)
-                //TODO: We can come back later to this when for instance we need BYTE_ARRAY for something else other than Image
+                //TODO we need a KLReadOnlyByteArrayControl, this doesn't exist yet
                 control = new KLReadOnlyImageControl();
             }
```
carldea, line 242: For byte array field we should create new `KlByteArrayFieldFactory()` and a `DefaultKLByteArrayControl` wrapping the `KLByteArrayControl`
We should have an else statement as a catch all using `KLReadOnlyDataTypeControl<T>`. We can work with @dukke . E.g. for digraph or ditree i basically to string the object.

Or another way is to use my original readonly factory:
`ReadOnlyKLFieldFactory rowf = ReadOnlyKLFieldFactory.getInstance();`

kview/src/main/java/dev/ikm/komet/kview/mvvm/model/DataModelHelper.java
```diff
@@ -104,25 +105,28 @@ public class DataModelHelper {
...
                 Entity.getFast(STRING.nid()),
                 Entity.getFast(COMPONENT_FIELD.nid()),
                 Entity.getFast(COMPONENT_ID_SET_FIELD.nid()),
                 Entity.getFast(COMPONENT_ID_LIST_FIELD.nid()),
-                Entity.getFast(DITREE_FIELD.nid()),
-                Entity.getFast(DIGRAPH_FIELD.nid()),
-                Entity.getFast(CONCEPT_FIELD.nid()),
-                Entity.getFast(SEMANTIC_FIELD_TYPE.nid()),
+//                Entity.getFast(DITREE_FIELD.nid()),
+//                Entity.getFast(DIGRAPH_FIELD.nid()),
```
carldea, line 114: Can we uncomment these field types?
we should and must have these field types. All Axioms are this type. User should be able to create Patterns. but when creating semantics (based on the pattern) both sides (readonly/left and editable/right) will be a readonly control the user cannot edit.

kview/src/main/java/dev/ikm/komet/kview/mvvm/model/DataModelHelper.java
```diff
@@ -104,25 +105,28 @@ public class DataModelHelper {
...
                 Entity.getFast(FLOAT_FIELD.nid()),
                 Entity.getFast(BOOLEAN_FIELD.nid()),
-                Entity.getFast(BYTE_ARRAY_FIELD.nid()),
-                Entity.getFast(ARRAY_FIELD.nid()),
-                Entity.getFast(INSTANT_LITERAL.nid()),
-                Entity.getFast(LONG.nid()),
-                Entity.getFast(VERTEX_FIELD.nid()),
+                //FIXME add byte array as its own type that is NOT an image
+//                Entity.getFast(BYTE_ARRAY_FIELD.nid()),
+                Entity.getFast(IMAGE_FIELD.nid())
```
carldea, line 122: Can we uncomment byte array field type?

dholubek: fixed

kview/src/main/java/dev/ikm/komet/kview/mvvm/model/DataModelHelper.java
dholubek: fixed

dholubek: > Can we uncomment these field types? we should and must have these field types. All Axioms are this type. User should be able to create Patterns. but when creating semantics (based on the pattern) both sides (readonly/left and editable/right) will be a readonly control the user cannot edit.

The only issue is that we need tickets to handle the either with read only string or with their own custom UI controls.  Handling them with read only strings will be a story that would also have to treat them as a non-participant in the form: no right click + edit, and not affect the submit button being enabled etc.  So there is more work associated than just displaying them.  Uncommenting in the mean time these leaves us open to defects being logged in the absence of that work.

dholubek: > > Can we uncomment these field types? we should and must have these field types. All Axioms are this type. User should be able to create Patterns. but when creating semantics (based on the pattern) both sides (readonly/left and editable/right) will be a readonly control the user cannot edit.
> 
> @carldea The only issue is that we need tickets to handle the either with read only string or with their own custom UI controls. Handling them with read only strings will be a story that would also have to treat them as a non-participant in the form: no right click + edit, and not affect the submit button being enabled etc. So there is more work associated than just displaying them. Uncommenting in the mean time leaves us open for defects being logged in the absence of that work.

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
dholubek: the catch all else is scoped for another story:
https://ikmdev.atlassian.net/browse/IIA-1912

dholubek: @carldea handling the Byte Array I think should be a different story, we don't have a KLReadOnlyByteArray currently.  The stories were for removing unsupported datatypes that cause exceptions and to map the image data type as a choice when creating the pattern fields.  I think it should be a sepearate effort to handle the Byte Array since there there is a left and right component so the scope will span more than this if/else block.

kview/src/main/java/dev/ikm/komet/kview/mvvm/model/DataModelHelper.java
dholubek: @carldea ok.  they are uncommented now and for the mean time.

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
```diff
@@ -83,6 +85,9 @@ public static Node generateNode(FieldRecord fieldRecord, ObservableField observa
         } else if (dataTypeNid == TinkarTerm.BOOLEAN_FIELD.nid()) {
             KlBooleanFieldFactory klBooleanFieldFactory = new KlBooleanFieldFactory();
             node = klBooleanFieldFactory.create(observableField, viewProperties.nodeView(), editable).klWidget();
+        } else if (dataTypeNid == TinkarTerm.IMAGE_FIELD.nid()) {
+            KlImageFieldFactory imageFieldFactory = new KlImageFieldFactory();
+            node = imageFieldFactory.create(observableField, viewProperties.nodeView(), editable).klWidget();
```
carldea, line 90: Not sure why GitHub doesn't allow me to add feedback to code lines below, but I have a question.
For Byte array field why is this code returning an image custom control?
Maybe this is older code.

kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldHelper.java
dukke: We also, at some point, still need to test the existing editable byte array control.. there are some stories for that in the backlog

dukke: @dholubek @carldea perhaps we could just remove this if statement for bytearray now? Since we don't really support it yet... (it was being used for Image before but now we don't need it for image or support byte array)

dukke (changes requested):

carldea (approved):
I'm going to approve so we can move this along. Looks good to me. We just have to be watchful if the data team decides to create data the UI doesn't support to throw a runtime exception.

dukke (approved):

PR #412, opened 2025-04-30
Feature/finished/iia 1892 submit other name causes exception in create mode

jdsmithsos:
Jira ticket:

https://ikmdev.atlassian.net/browse/IIA-1892

The problem is actually in DetailsController.generateOtherNameRow(DescrName) and is caused by the MODE still being in CREATE.

During CREATE mode the otherName semanticId has not been set.

Only when the Concept goes into edit mode is the semanticId set, which is done within ConceptViewModel.createConcept().  But only if there is an Axiom defined in the Concept.

In CREATE mode the Fully Qualified Name and Other Names are stored in the ConceptViewModel, they have not yet been persisted.

When an Axiom exists and an Other Name is created, only then does the FQN and Other Name persist, and the other name’s sematicId is set.

Adding a conditional in generateOtherNameRow() for a null semanticId to use the other name as entered rather than using the semanticId to get the name fixes the problem.

image

After fix:

image

Note that the Concept is still in CREATE mode.  An Axiom has not yet been created.

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
@@ -15,37 +15,6 @@
  */
 package dev.ikm.komet.kview.mvvm.view.details;
 
-import static dev.ikm.komet.framework.events.FrameworkTopics.RULES_TOPIC;
-import static dev.ikm.komet.kview.fxutils.IconsHelper.IconType.ATTACHMENT;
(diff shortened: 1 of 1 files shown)


jdsmithsos:
> Looks good to me. Make sure you've tested the creation of the new Concept contain the new Other name added.

I tested these two scenarios with this change:

- Create FQN then Axiom, which puts the Concept into EDIT mode, then create Other Name
- Create FQN then Other Name, which Concept will still be in CREATE mode, then create Axiom, which goes to EDIT mode

PR #421, opened 2025-05-05
Feature/finished/iia 1599 view calculator refresh issue after import

carldea:
After an import, the code classic Concept Navigator and Window (summoned inside of a Journal view will show the latest from the import process.

### Before import
Screenshot 2025-05-04 at 11 44 32 PM

### After import
Screenshot 2025-05-04 at 11 45 34 PM

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/model/ViewCoordinateHelper.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/model/ViewCoordinateHelper.java
@@ -29,7 +29,7 @@ public class ViewCoordinateHelper {
      */
     public static void changeViewCalculatorToLatestByTime(ViewProperties viewProperties, long time) {
         ViewCalculator existingViewCalculator = viewProperties.calculator();
-        StampCoordinateRecord latestStampCoordinate = existingViewCalculator.vertexStampCalculator().filter().withStampPositionTime(time);
+        StampCoordinateRecord latestStampCoordinate = existingViewCalculator.stampCoordinate().withStampPositionTime(time).toStampCoordinateRecord();
         ViewCoordinateRecord latestViewCoordinate = existingViewCalculator.viewCoordinateRecord().withStampCoordinate(latestStampCoordinate);
         Platform.runLater(() -> viewProperties.nodeView().setValue(latestViewCoordinate));
     }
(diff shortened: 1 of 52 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/model/ViewCoordinateHelper.java
```diff
@@ -0,0 +1,94 @@
...
+public class ViewCoordinateHelper {
+    private static final Logger LOG = LoggerFactory.getLogger(ViewCoordinateHelper.class);
+    /**
+     * TODO: Experimental - will change view coordinates viewProperties.nodeView().setValue(latestViewCoordinate));
+     * Changes the current view coordinate with the current datetime, thus updating its view calculator.
+     * @param viewProperties A given view property
+     */
+    public static void changeViewCalculatorToLatestByTime(ViewProperties viewProperties, long time) {
+        ViewCalculator existingViewCalculator = viewProperties.calculator();
+        StampCoordinateRecord latestStampCoordinate = existingViewCalculator.vertexStampCalculator().filter().withStampPositionTime(time);
```
aks8m, line 32: What is the reason for calling vertexStampCalculator() here? It seems to be coming from the NavigationCalculatorDelegate but is just returning a StampCalculatorWithCache. Could we get a StampCalculatorWithCache somewhere else? I see that it really is only in the NavigationCalculatorWithCache but what about the ViewCalculatorWithCach class?

kview/src/main/java/dev/ikm/komet/kview/mvvm/model/ViewCoordinateHelper.java
```diff
@@ -0,0 +1,94 @@
...
+        Platform.runLater(() -> viewProperties.nodeView().setValue(latestViewCoordinate));
+    }
+
+    /**
+     * Create a new View Calculator based on the current datetime + 24 hours with an updated view coordinate & view calculator.
+     * Note: This appears to work however the records returned could be uncommitted records as they have a larger latest date.
+     * @param viewProperties A given view property
+     */
+    public static ViewCalculatorWithCache createViewCalculatorLatestByTime(ViewProperties viewProperties) {
+        return createViewCalculatorLatestByTime(viewProperties, System.currentTimeMillis() + (1000 * 60 * 60 * 24)); // 1ms x 60s x 60m x 24hrs
```
aks8m, line 43: Does the + 24 do something that currentTImeMillis() isn't. Help me understand the need for an additional 24 hours to current time when retrieving Latest.

aks8m (approved):
Looks good. I

PR #422, opened 2025-05-05
Use isDefined API

jperedadnr:
instead of `hasSufficientSet`, to set if the concept in the ConceptNavigator control is defined or primitive.


carldea (approved):
Looks good to me.

PR #423, opened 2025-05-05
Feature/finished/iia 1529 top drop line reworked

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-1529

Reworked the solution which includes:

- shared method to create the dropLine for the topDropLine and the components
- the problem that was discovered before the demo was caused by the width of the drop line being the same width as the contentWidth, which caused a recursive call to layoutChildren().  This is fixed by subtracting at least 1 from the drop line width.

Before this rework the widths of the component controls grew by 2 pixels indefinitely.

After this fix:

During drag the component widths remain the same, rather than increasing

image

Drag to the top

image

After the drag all components are the same width, which didn't increase

image


dukke (changes requested):
@jdsmithsos 
Just remembered, padding is not enough...
I see there's a bug in this skin.
We should do getInsets() rather than do getPadding(). Getting the insets also will return to us the border + padding, and if we do getPadding() we only get the padding. 
So, we might not need this magic numbers (-2).

carldea (approved):
Look good to me. Let's make sure the comments describe the magic number and Insets vs Padding.

jdsmithsos:
Changes done as request for approving PR for merge.

dukke (approved):
Good work! It's nice that you were able to centralize the logic for drawing a drop line into just 1 method!

PR #424, opened 2025-05-05
Feature/finished/iia 1920 float integer control skin exceptions

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-1920

Summary of changes:

- added optional minus sign at beginning of KLIntegerControlSkin regex:  "^-?(0|[1-9][0-9]*)$"
- restructured conditionals to allow for additional checks of "-", "+", and exponent
- added checks for "-" (and "+" since float regex allows optional + at beginning) in KLFloatControlSkin

Before fix:

For Integer field, enter - followed by a decimal

image

For Float field, enter -

image

After fix:

Integer field now allows - and decimal without error

image

Float field now no exception:

image

image

image


dholubek (approved):
looks good

PR #432, opened 2025-05-07
Feature/finished/iia 1941 bugfix nodes index

dholubek:
This fixes an exception being thrown when the nodes collection is empty.  We have a placeholder image on tinkar test pattern 3 and this fixes that exception.  You are allowed to create a pattern with zero fields today in Komet, so I this check should exist.  This also fixes an exception that I found in https://ikmdev.atlassian.net/browse/IIA-1933


bhharsh13 (approved):
Looks good to me

PR #433, opened 2025-05-07
Feature/finished/iia 1512 autosave with debouncing for string integer float

swaroopsalvi:


dholubek (changes requested):
@swaroopsalvi  This PR says it has 37 commits.  I faced a similar issue when I had commits in my local main by accident that I never intended to get into ikmdev/komet main.  Your forked repo says "36 commits ahead, 6 commits behind".  You are going to potentially run into the same issue that I did where git will eventually not reconcile a common ancestor properly and start bringing back old commits.  If not fixed over time there is a good chance we will start to see zombie defects come back.  

The fix is to delete your forked repo and re-fork a completely clean  repo.

PR #434, opened 2025-05-08
Feature/finished/iia 1599 viewcalc refresh issue 2

carldea:
I did more testing and the new code to derive a new view property derived from the concept navigator throws an exception during the create concept window.


dholubek (approved):
looks good

PR #437, opened 2025-05-08
Allow horizontal scrolling for LineageBox

jperedadnr:
This PR allows scrolling horizontally so that the full name of the concept can be seen in the lineage box.


carldea (approved):
Looks good to me!

PR #439, opened 2025-05-08
Code to delay autosave.

swaroopsalvi:


kview/src/main/java/dev/ikm/komet/kview/klfields/floatfield/DefaultKlFloatField.java
```diff
@@ -17,8 +19,14 @@ public DefaultKlFloatField(ObservableField<Float> observableFloatField, Observab
...
-            floatControl.valueProperty().bindBidirectional(observableFloatField.valueProperty());
-            floatControl.setTitle(getTitle());
+            floatControl.valueProperty().set(observableFloatField.valueProperty().getValue());
+            Timeline timeline = createTimeline();
+            timeline.setOnFinished((evt) -> {
+                observableFloatField.valueProperty().setValue(floatControl.getValue());
+            });
+            floatControl.valueProperty().subscribe( s -> {
+                timeline.playFromStart();
+            });
```
carldea, line 29: Now if the bidirection is missing how would the UI get updated from observable's valueproperty?
- bidirectional means changes from user go to backend, and visaversa UI control <----> Observable field
- this is only unidirectional UI control -----> observable field.

PR #440, opened 2025-05-08
Feature/finished/iia 1933 last commit vals

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1933

the observable fields that send over the event bus will revert (all except the string on Tinkar example 1 data) to their default values after the first iteration of the for loop in GenEditingDetailsController when updating the left side's real observable fields.  I changed this to send plain objects instead.

edit window before editing:
edit window before editing

edit window before submit:
edit window before submit

edit window after submit:
edit window after submit

re-summon window after submit:
re-summon window after submit

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
@@ -14,11 +14,11 @@ public class GenEditingEvent extends Evt {
 
     public static final EvtType<GenEditingEvent> VERSION_UPDATED = new EvtType<>(Evt.ANY, "VERSION_UPDATED");
 
-    private List<ObservableField<?>> list;
+    private List<?> list;
 
     private int nid;
 
-    public GenEditingEvent(Object source,EvtType eventType, List<ObservableField<?>> list, int nid){
+    public GenEditingEvent(Object source,EvtType eventType, List<?> list, int nid){
         super(source,eventType);
         this.list = list;
         this.nid = nid;
@@ -28,7 +28,7 @@ public class GenEditingEvent extends Evt {
         super(source,eventType);
     }
 
-    public List<ObservableField<?>> getList(){
+    public List<?> getList(){
         return list;
     }
 
--- a/kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
@@ -14,11 +14,11 @@ public class GenEditingEvent extends Evt {
 
     public static final EvtType<GenEditingEvent> VERSION_UPDATED = new EvtType<>(Evt.ANY, "VERSION_UPDATED");
 
-    private List<?> list;
+    private List<Object> list;
 
     private int nid;
 
-    public GenEditingEvent(Object source,EvtType eventType, List<?> list, int nid){
+    public GenEditingEvent(Object source,EvtType eventType, List<Object> list, int nid){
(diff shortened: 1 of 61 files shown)


kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
```diff
@@ -14,11 +14,11 @@ public class GenEditingEvent extends Evt {
 
     public static final EvtType<GenEditingEvent> VERSION_UPDATED = new EvtType<>(Evt.ANY, "VERSION_UPDATED");
 
-    private List<ObservableField<?>> list;
+    private List<?> list;
 
     private int nid;
 
-    public GenEditingEvent(Object source,EvtType eventType, List<ObservableField<?>> list, int nid){
+    public GenEditingEvent(Object source,EvtType eventType, List<?> list, int nid){
```
carldea, line 21: Should the list be this: `List<Object>`?
Using `?` wildcard still means a specific type. `Object` means any kind of object.

dholubek: fixed

kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
```diff
@@ -14,11 +14,11 @@ public class GenEditingEvent extends Evt {
...
     public static final EvtType<GenEditingEvent> VERSION_UPDATED = new EvtType<>(Evt.ANY, "VERSION_UPDATED");
 
-    private List<ObservableField<?>> list;
+    private List<Object> list;
 
     private int nid;
 
-    public GenEditingEvent(Object source,EvtType eventType, List<ObservableField<?>> list, int nid){
+    public GenEditingEvent(Object source,EvtType eventType, List<Object> list, int nid){
         super(source,eventType);
```
swaroopsalvi, line 22: the list we are setting is always an ObservableField so why did we change to object ?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
```diff
@@ -328,8 +328,10 @@ private void resetFieldValues(EntityVersion entityVersion) {
     @FXML
     public void submit(ActionEvent actionEvent) {
        cancelButton.requestFocus();
+
        //create new list for passing to the event.
-       List<ObservableField<?>> list = new ArrayList<>(observableFields);
+       List<Object> list = new ArrayList<>(observableFields.size());
+       observableFields.forEach(observableField -> list.add(observableField.value()));
```
swaroopsalvi, line 335: Why can't we keep it as ObservableField ?

swaroopsalvi (commented):
The reset / cancel / clear functionality is not required.

kview/src/main/java/dev/ikm/komet/kview/events/genediting/GenEditingEvent.java
dholubek: the observable fields change and revert to previous values on the second iteration of the for loop.  we don't need to put an observable field on the event bus since we just want to pass the information.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
dholubek: @swaroopsalvi  see above comment.

carldea (approved):
Looks good to me!
Thanks @dholubek

PR #443, opened 2025-05-09
Feature/finished/iia 1928 style type ahead popup

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1928

image


carldea (approved):
Looks good to me!
Thank you for all this hard work. Looking better and better. 
:-)

PR #444, opened 2025-05-09
Feature/finished/iia 1955 concurrent modification exception calling listeners

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-1955

Summary of changes:

The problem is within ObservableCoordinateAbstract.setValue() and changeBaseCoordinate().

setValue() is being called recursively.  See stack trace below.

This is specifically happening in ObservableLanguageCoordinateNoOverride, which extends ObservableCoordinateAbstract.

An ArrayList cannot be modified when it is being iterated.

A fix for this issue is to put the for loop iterator within Platform.runLater() so each recursive iteration will be queued in the JavaFX Application Thread.

Before fix:

image

After fix, no exception dialog:

With Fully Qualified Name selected

image

With Regular Name selected

image

Other bugs exist, which will be documented by Sundeep.


dholubek (approved):
looks good

swaroopsalvi (approved):
Looks good to me.

PR #445, opened 2025-05-09
Add enum with actions and change onAction API to allow different consumers

jperedadnr:
This PR allows the user to open a concept Window by selecting the Open in Workspace action of the ContextMenu that shows up when a a single selection is done in the concept navigator control.

But it also redefines the `KLConceptNavigatorControl::onActionProperty` API, by adding an enum with the list of possible actions from the context menus that are shown when a single or multiple selection are done in the control.

For each action, the developer that uses this control has to provide a consumer of the item(s) affected by the selection, like opening it/them in the workspace.

For the related concepts action, a new API has been added to the ConceptNavigatorTreeItem, so each item can have a list of related concepts. This should be filled when the treeItem is created, for instance, or via Navigator (TODO).

### DIFF
--- a/application/src/test/java/dev/ikm/komet/sampler/controllers/SamplerConceptNavigatorController.java
+++ b/application/src/test/java/dev/ikm/komet/sampler/controllers/SamplerConceptNavigatorController.java
@@ -34,6 +34,8 @@ import javafx.scene.control.TreeItem;
 import javafx.scene.input.Dragboard;
 import javafx.scene.input.TransferMode;
 import javafx.scene.layout.VBox;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -47,6 +49,8 @@ import static dev.ikm.komet.kview.controls.KLConceptNavigatorTreeCell.CONCEPT_NA
 import static dev.ikm.komet.preferences.JournalWindowPreferences.MAIN_KOMET_WINDOW;
 
 public class SamplerConceptNavigatorController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SamplerConceptNavigatorController.class);
 
     @FXML
     private Label samplerDescription;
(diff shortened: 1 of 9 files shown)


application/src/test/java/dev/ikm/komet/sampler/controllers/SamplerConceptNavigatorController.java
```diff
@@ -189,22 +197,39 @@ public void initialize() {
...
+            case SHOW_RELATED_CONCEPTS -> {
+                // Dummy, for now just add the parents of the selected item as related content:
+                TreeItem<ConceptFacade> selectedItem = conceptNavigatorControl.getSelectionModel().getSelectedItem();
+                if (selectedItem != null) {
+                    conceptNavigatorControl.getNavigator().getParentNids(selectedItem.getValue().nid());
+                    List<ConceptFacade> list = Arrays.stream(conceptNavigatorControl.getNavigator().getParentNids(selectedItem.getValue().nid())).boxed()
+                            .map(nid -> (ConceptFacade) Entity.getFast(nid)).toList();
+                    ((ConceptNavigatorTreeItem) selectedItem).setRelatedConcepts(list);
+                }
+                yield i -> System.out.println("Click! on " + i.description());
```
carldea, line 215: Can we add Logger instead?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
```diff
@@ -959,10 +961,30 @@ public void loadConceptNavigatorPanel() {
...
+            case SHOW_RELATED_CONCEPTS -> {
+                // Dummy, for now just add the parents of the selected item as related content:
+                TreeItem<ConceptFacade> selectedItem = conceptNavigatorControl.getSelectionModel().getSelectedItem();
+                if (selectedItem != null) {
+                    conceptNavigatorControl.getNavigator().getParentNids(selectedItem.getValue().nid());
+                    List<ConceptFacade> list = Arrays.stream(conceptNavigatorControl.getNavigator().getParentNids(selectedItem.getValue().nid())).boxed()
+                            .map(nid -> (ConceptFacade) Entity.getFast(nid)).toList();
+                    ((ConceptNavigatorTreeItem) selectedItem).setRelatedConcepts(list);
+                }
+                yield i -> System.out.println("Click! on " + i.description());
```
carldea, line 979: Can we use a logger to output to the console?

application/src/test/java/dev/ikm/komet/sampler/controllers/SamplerConceptNavigatorController.java
jperedadnr: done

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/journal/JournalController.java
jperedadnr: done

carldea (approved):
@jperedadnr Excellent work! 
Now it is much more flexible for the user of the API.

PR #446, opened 2025-05-09
Feature/finished/iia 1970 type ahead reference component

dukke:
Fixes: 
- https://ikmdev.atlassian.net/browse/IIA-1970
- https://ikmdev.atlassian.net/browse/IIA-1723

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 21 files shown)


carldea (approved):
Looks good to me!
The user will be more productive.
 ;-)

dukke:
> Looks good to me! The user will be more productive. ;-)

Thanks @carldea ! Your tip on calling toProxy() was very valuable back when I was creating the AutoCompleteTextField and needed to use an EntityProxy rather than a ConceptFacade.

PR #450, opened 2025-05-12
Feature/finished/iia 1324 field name from meaning

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1324

disabling the user from entering the display name and deriving this from the meaning.description() text

image


carldea (approved):
looks good to me!

PR #454, opened 2025-05-12
Add support for drag and drop concepts into the concept navigator

jperedadnr:
Based on the same mechanism used to expand the treeView when hovering/selecting search results.


carldea (approved):
Looks good to me.

PR #457, opened 2025-05-13
IIA-1965-Exception-Triggered-When-Submitting-Semantic-After-Removing-Concept-from-CSET-CLIST

swaroopsalvi:


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
```diff
@@ -350,9 +350,6 @@ public void submit(ActionEvent actionEvent) {
                // EventBus implementation changes to refresh the details area if commit successful
                EvtBusFactory.getDefaultEvtBus().publish(genEditingViewModel.getPropertyValue(CURRENT_JOURNAL_WINDOW_TOPIC),
                        new GenEditingEvent(actionEvent.getSource(), PUBLISH, list, semantic.nid()));
-               // refesh the pattern navigation
-               EvtBusFactory.getDefaultEvtBus().publish(SAVE_PATTERN_TOPIC,
-                       new PatternSavedEvent(actionEvent.getSource(), PatternSavedEvent.PATTERN_UPDATE_EVENT));
            }, () -> {
```
dholubek, line 353: why are we commenting this out?  this will create a defect where the semantic won't get refreshed in the pattern navigator.

dholubek (changes requested):
see comment about pattern update event

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
dholubek: @swaroopsalvi  explained they are refreshed by being observables... closing review

PR #458, opened 2025-05-13
Feature/finished/iia 1992 1967 change language coord starburst

jdsmithsos:
Jira tickets:

https://ikmdev.atlassian.net/browse/IIA-1992
https://ikmdev.atlassian.net/browse/IIA-1967

These defects are related, both with the Change description preference menu item.

Summary of changes:

- ObservableViewWithOverride.makeLanguageCoordinateListProperty() - this method was directly using the ObservableLanguageCoordinateNoOverride, rather than converting to ObservableLanguageCoordinateWithOverride.  The starburst icon only changes to the orange color when overrides have been set, and the WithOverride class was not being used (at all, anywhere in the code).  Added conversion to the WithOverride class just like all of the other make* methods do.
- ViewMenuTask.addChangeItemsForView() - modified to check if there are multiple languages, and if not, now doesn't append the index number of the language for the Menu text.
- ViewMenuTask refactored to have a common shared method createChangeDescriptionMenuItems(), which is used in both addChangeItemsForView() and addChangeItemsForLanguage()
- ViewMenuTask.changeStates() - fixed the issue of the CheckMenuItem selected indicator (the check) not changing when menu items are selected
- Testing with Classic Komet discovered that there is a ConcurrentModificationException on Classic initialization.  This seems to have been caused by the converting to ObservableLanguageCoordinateWithOverride, but not directly caused by.  Modified the ObservableCoordinateAbstract.setValue() to copy the listener list and to use the copy to call the listeners.

Before changes:

There is a 0 appended in the Menu text, and the language coordinate Change description preference selected is Regular name.
Note that the Navigator does not have any "SOLOR" in the names, which is only available for Fully Qualified Name.

image

When Fully Qualify Name is selected, SOLOR is in the navigator text.  But the starburst icon is still grey, not orange.  The starburst icon only changes to orange color when there are overrides.

image

Change allowed states, initially with default of ACTIVE, INACTIVE.  Before selecting ACTIVE.

image

Change allowed states, after selecting ACTIVE.  Note that the ACTIVE CheckMenuItem is not selected with a check.

image

After changes:

Since there is only one language, there is now no longer an appended index in the Change language coordinate menu text.

image

When the Fully Qualified Name Change description preference is selected, the starburst icon is now orange because the ObservableLanguageCoordinateWithOverride class is being used.

image

Change allowed states, after selecting ACTIVE.  ACTIVE is now checked.

image

### DIFF
--- a/framework/src/main/java/dev/ikm/komet/framework/view/ObservableCoordinateAbstract.java
+++ b/framework/src/main/java/dev/ikm/komet/framework/view/ObservableCoordinateAbstract.java
@@ -16,7 +16,6 @@
 package dev.ikm.komet.framework.view;
 
 import dev.ikm.tinkar.coordinate.ImmutableCoordinate;
-import javafx.application.Platform;
 import javafx.beans.InvalidationListener;
 import javafx.beans.property.Property;
 import javafx.beans.property.SimpleObjectProperty;
@@ -26,7 +25,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
-import java.util.ConcurrentModificationException;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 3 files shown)


dholubek (approved):
looks good... I'm not perfectly familiar with changing some of these things in classic Komet since I have done it so rarely.  I might need to defer to some on the architecture team @aks8m @rbradley813

carldea (approved):
This looks good to me.
Glad we could also fix or improve avoiding concurrency mod exception.

PR #460, opened 2025-05-13
Feature/finished/iia 2000 exception editing new pattern

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-2000

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.51.0-IIA-2000-Exception-Editing-New-Pattern-SNAPSHOT</version>
+        <version>1.51.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.51.0-SNAPSHOT</version>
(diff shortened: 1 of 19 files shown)


swaroopsalvi (changes requested):
We do not need to call the `toProxy()` method on entity service.
The ViewModel is using EntityFacade, so no need to call toProxy on it.

I will suggest the fix to be as below:
  ` definitionViewModel.setPropertyValue(PURPOSE_ENTITY, patternDefinition.purpose());
                    definitionViewModel.setPropertyValue(MEANING_ENTITY, patternDefinition.meaning());
`
You will have to also change the way you are getting the description from the meaning in `PatternFieldsController.java`
So something like: 
`   ConceptEntity conceptEntity = Entity.getFast((EntityFacade) meaningObject);
                displayNameTextField.setText(conceptEntity.description());
`

dukke:
> We do not need to call the `toProxy()` method on entity service. The ViewModel is using EntityFacade, so no need to call toProxy on it.
> 
> I will suggest the fix to be as below: `definitionViewModel.setPropertyValue(PURPOSE_ENTITY, patternDefinition.purpose()); definitionViewModel.setPropertyValue(MEANING_ENTITY, patternDefinition.meaning());` You will have to also change the way you are getting the description from the meaning in `PatternFieldsController.java` So something like: ` ConceptEntity conceptEntity = Entity.getFast((EntityFacade) meaningObject); displayNameTextField.setText(conceptEntity.description());`

Hi @swaroopsalvi , you're referring to the old code however that has changed recently. We now need to use toProxy() because our component controls require an EntityProxy.
This PR is not about using toProxy(), i.e. that's not what this PR changes. toProxy() was already there and working however it can be null so in this PR I'm guarding against that possibility so that we don't get a NPE.
Thanks!

dholubek (approved):
approved

swaroopsalvi (approved):

PR #464, opened 2025-05-14
IIA-1971 New Path added in STAMP Module

bhharsh13:
Fixed the issue for the new Path added in the STAMP Module. 
https://ikmdev.atlassian.net/browse/IIA-1971
New Path Added in module

Please help to review the PR.

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.51.0-IIA-1971-new-path-stamp-module-SNAPSHOT</version>
+        <version>1.51.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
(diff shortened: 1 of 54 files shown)


swaroopsalvi (changes requested):
Can you also update the following methods in StampVeiwModel to call the fetchDescendentsOfConcets? 
1. findAllModules
2. findAllPaths

So the body of the methods will be:

public Set<ConceptEntity> findAllModules(ViewProperties viewProperties) {
        return DataModelHelper.fetchDescendentsOfConcept(viewProperties, TinkarTerm.MODULE.publicId());
    }

public Set<ConceptEntity> findAllPaths(ViewProperties viewProperties) {
        return DataModelHelper.fetchDescendentsOfConcept(viewProperties, TinkarTerm.PATH.publicId());
    }

You will have to test all the screens including LIDR.

@carldea , This will have consistent implementation across all the dropdowns.
Please suggest.

swaroopsalvi (approved):
Looks good. Need to do unit testing for the modules that call these methods.

PR #465, opened 2025-05-14
Feature/finished/iia 1952 edit coord concept in journal

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1952

existing coordinate menu in the classic Komet Concept navigator
image

we are showing this with from the bulls-eye onward to match the bulls-eye that is in the Concept's window title bar
**NOTE the menu is display only**
image

apologies for the large diff on concept-details.fxml... this appears to be SceneBuilder formatting on save


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
```diff
@@ -346,8 +358,9 @@ public class DetailsController  {
     public DetailsController() {
     }
 
-    public DetailsController(UUID conceptTopic) {
+    public DetailsController(UUID conceptTopic, ViewProperties viewProperties) {
```
carldea, line 361: Can we pass item through the view model?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsNode.java
```diff
@@ -100,7 +99,7 @@ private void init(boolean displayOnJournalView) {
             // 2) not in fxml view class    - apply(file, view, ...view models)
             // 3) not in fxml view instance - apply(file, view instance, ...view models)
             Config config = new Config(getClass().getResource(CONCEPT_DETAILS_VIEW_FXML_FILE))
-                    .controller(new DetailsController(conceptTopic))
+                    .controller(new DetailsController(conceptTopic, viewProperties))
                     .updateViewModel("conceptViewModel", viewModel ->
                             viewModel.setPropertyValue(CURRENT_JOURNAL_WINDOW_TOPIC, journalWindowTopic));
```
carldea, line 104: Is there a difference between the `conceptTopic` & `journalWindowTopic`?

I prefer controllers to be an empty constructor. JavaFX's makes values available in the `initialize()` method.

dholubek: If I recall correctly the concept topic is for the concept window only and for things that only affect the current concept window like editing its FQN, we don't want all open concept windows to listen for that.  And the journal topic is for events inside the entire journal window like saving the concept itself so that a navigator can be refreshed.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
dholubek: yes... I will update it...

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsNode.java
carldea: Makes sense. 
I think the data passing (to constructor) was before Cognitive apis. We can pass in 'conceptTopic and viewProperties' in the updateViewModel() method. 

Not sure but check and see if we need to call updateView( ... ). Another older mechanism to pass data to update the controller or view.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
dholubek: @carldea actually can we do this refactoring after the demo?  it affected the navigation from the description pencil icon

carldea: yes. Please create a new ticket to refactor (clean up).
I will approve.

carldea (approved):
Looks good to me.

PR #466, opened 2025-05-15
Feature/finished/jpro 131 132 133 changeset exchange using GitHub account

besidev:
This PR solution spans over 3 Jira tickets:

**JIRA Ticket**: [JPRO-131](https://ikmdev.atlassian.net/browse/JPRO-131)
**JIRA Ticket**: [JPRO-132](https://ikmdev.atlassian.net/browse/JPRO-132)
**JIRA Ticket**: [JPRO-133](https://ikmdev.atlassian.net/browse/JPRO-133)

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright © 2015 Integrated Knowledge Management (support@ikm.dev)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.ikm.komet.kview.mvvm.view.changeset.exchange;
+
+import com.jpro.webapi.WebAPI;
+import dev.ikm.komet.kview.mvvm.viewmodel.GitHubPreferencesViewModel;
+import dev.ikm.komet.preferences.KometPreferences;
+import dev.ikm.komet.preferences.Preferences;
+import javafx.beans.value.ChangeListener;
+import javafx.event.ActionEvent;
+import javafx.fxml.FXML;
+import javafx.fxml.Initializable;
+import javafx.scene.control.Button;
+import javafx.scene.control.Hyperlink;
+import javafx.scene.control.PasswordField;
+import javafx.scene.control.TextField;
+import javafx.scene.text.Text;
+import one.jpro.platform.utils.OpenLink;
+import org.carlfx.cognitive.loader.InjectViewModel;
+import org.carlfx.cognitive.validator.ValidationMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URL;
+import java.util.ResourceBundle;
+import java.util.prefs.BackingStoreException;
+import java.util.stream.Stream;
+
+import static dev.ikm.komet.kview.mvvm.view.changeset.exchange.GitPropertyName.*;
+
+/**
+ * Controller class for handling GitHub preferences in the application.
+ * <p>
+ * This controller manages the UI components and user interactions for configuring
+ * GitHub repository settings. It validates input fields, displays error messages,
+ * and persists GitHub connection information to the application preferences.
+ * <p>
+ * The controller follows the MVVM (Model-View-ViewModel) pattern, with field validation
+ * and business logic delegated to the associated {@link GitHubPreferencesViewModel}.
+ *
+ * @see GitHubPreferencesViewModel
+ * @see KometPreferences
+ */
+public class GitHubPreferencesController implements Initializable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GitHubPreferencesController.class);
+
+    @FXML
+    private TextField gitUrlTextField;
+
+    @FXML
+    private TextField gitEmailTextField;
+
+    @FXML
+    private TextField gitUsernameTextField;
+
+    @FXML
+    private PasswordField gitPasswordField;
+
+    @FXML
+    private Text errorText;
+
+    @FXML
+    private Hyperlink signUpHyperlink;
+
+    @FXML
+    private Button cancelButton;
+
+    @FXML
+    private Button connectButton;
+
+    @InjectViewModel
+    private GitHubPreferencesViewModel githubPreferencesViewModel;
+
+    /**
+     * Initializes the controller class.
+     *
+     * @param location  The location used to resolve relative paths for the root object, or {@code null} if not known.
+     * @param resources The resources used to localize the root object, or {@code null} if not localized.
+     */
+    @Override
+    public void initialize(URL location, ResourceBundle resources) {
+        ChangeListener<String> isNotPopulatedListener = (observable, oldValue, newValue) ->
+                githubPreferencesViewModel.getValidators(IS_NOT_POPULATED).stream()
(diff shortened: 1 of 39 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
```diff
@@ -0,0 +1,276 @@
...
+    public void initialize(URL location, ResourceBundle resources) {
+        ChangeListener<String> isNotPopulatedListener = (observable, oldValue, newValue) ->
+                githubPreferencesViewModel.getValidators(IS_NOT_POPULATED).stream()
+                        .findAny()
+                        .ifPresent(validator -> validator.apply(null, githubPreferencesViewModel));
+
+        gitUrlTextField.textProperty().addListener(isNotPopulatedListener);
+        gitEmailTextField.textProperty().addListener(isNotPopulatedListener);
+        gitUsernameTextField.textProperty().addListener(isNotPopulatedListener);
+        gitPasswordField.textProperty().addListener(isNotPopulatedListener);
```
carldea, line 104: I was wondering could the above use this facility from the MVVM framework? 
Not exactly sure if it is the desired effect. 
https://github.com/carldea/cognitive/wiki#viewmodels-doonchange
```java
// fields change fire a validation
githubPreferencesViewModel.doOnChange(()->{
       githubPreferencesViewModel.validate();
}, GIT_URL, GIT_EMAIL, GIT_USERNAME, GIT_PASSWORD, CONNECT_BUTTON_STATE);

...
githubPreferencesViewModel.validateOnChange(IS_NOT_POPULATED, (validationMessages) -> {
    // process each message owned by Field
    validationMessages.forEach(validationMessage -> {
       // any fields 
       String message = validationMessage.interpolate(githubPreferencesViewModel);
        // apply to the UI
    });
});

```
Let me know if you need help. Really much of Cognitive is optional and all depends on personal preference.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
```diff
@@ -0,0 +1,276 @@
...
+     *   <li>Persists valid settings to application user preferences</li>
+     * </ul>
+     *
+     * @param actionEvent The event that triggered this handler
+     */
+    @FXML
+    public void handleConnectButtonEvent(ActionEvent actionEvent) {
+        githubPreferencesViewModel.save();
+
+        if (githubPreferencesViewModel.hasErrorMsgs()) {
```
carldea, line 132: **(Purely Optional)**
While this is fine, but I have revised some of the [demo code](https://github.com/carldea/cognitive/blob/main/src/test/java/org/carlfx/cognitive/test/demo/AccountCreateController.java#L88-L103). You might want to look at the way I handle errors when a single field changes (not all validators get fired). e.g. `validateOnChange()` is used so not all validators will evaluate but just the ones related to this field. 
Caveat... even my code still needs to be cleaner.  

view model's using validation view model can also make use of `.validProperty()` and `.invalidProperty()`. These are read-only properties. These are usually triggered when save() or validate() is called. Or also `.doOnChange()` as any field changes should call `.validate()`

carldea: Was thinking could preferences (persistent) logic be in another DAO type class or temporarily inside the ViewModel (possibly override save() method). after a valid save ui property values are copied to the value model layer. Just a thought. There's no hard and fast rule, just different ways to decouple business or presentation logic.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
besidev: Using `doOnChange` from the Cognitive API to set up property change listeners is a game changer, Carl. The code is way more minimal and readable.
We also tried to use the `githubPreferencesViewModel.validateOnChange` call:
```java
githubPreferencesViewModel.validateOnChange(IS_NOT_POPULATED, (validationMessages) -> {
    // process each message owned by Field
    validationMessages.forEach(validationMessage -> {
       // any fields 
       String message = validationMessage.interpolate(githubPreferencesViewModel);
        // apply to the UI
    });
});`
```
However, in this case the Cognitive API complained during runtime since `IS_NOT_POPULATED` is not a property - it was setup as a validator in the respective ViewModel class.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
besidev: > **(Purely Optional)** While this is fine, but I have revised some of the [demo code](https://github.com/carldea/cognitive/blob/main/src/test/java/org/carlfx/cognitive/test/demo/AccountCreateController.java#L88-L103). You might want to look at the way I handle errors when a single field changes (not all validators get fired). e.g. `validateOnChange()` is used so not all validators will evaluate but just the ones related to this field. Caveat... even my code still needs to be cleaner.
> 
> view model's using validation view model can also make use of `.validProperty()` and `.invalidProperty()`. These are read-only properties. These are usually triggered when save() or validate() is called. Or also `.doOnChange()` as any field changes should call `.validate()`

In our previous implementation, we consolidated all validation errors into a single paragraph (our own decision actually, as currently there are no specific guidelines for validation error handling in the UI). Thanks to your suggestion Carl, we've found that `validateOnChange()` provides an ideal solution for our needs.
This approach enables us to:

- Validate each input text field independently (the validators are now fired for each field separately on a change)
- Display specific, contextual error messages directly below the relevant field

While this implementation requires minor design modifications to accommodate individual error labels beneath each text field, the tradeoff significantly enhances user experience by creating a more intuitive and user-friendly validation flow.

besidev: > Was thinking could preferences (persistent) logic be in another DAO type class or temporarily inside the ViewModel (possibly override save() method). after a valid save ui property values are copied to the value model layer. Just a thought. There's no hard and fast rule, just different ways to decouple business or presentation logic.

That's an excellent suggestion, Carl. Moving the preferences persistence logic to a dedicated DAO-type class significantly improves our separation of concerns, creating cleaner boundaries between business and presentation layers. We've already implemented and integrated this approach in our latest changes.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/exchange/GitHubPreferencesController.java
carldea: @besidev @indritbeqiri, 
So glad to see these validation techniques and strategies being applied. API design is hard and UI Forms have different behaviors such as when the user first sees the screen as blank fields. but you don't want the user to see all the errors because all fields are required, but only when they type. Or maybe you do? (UX design preference)

Thank you for being the guinea pig using Cognitive MVVM! I still have improvements when time permits. I believe it allows JavaFX UI developers to have cleaner separation and allows them to be productive. I do think MVVM can be overkill at times, but for large projects it is a standardize way to develop/test and organize presentation and business logic without having to have everything loaded live (fxml, controller, scene). Tests can be smaller and more focused. The goal is your controller should be thin and initialize() is where wiring(binding) controls is done once. But again, there is no hard fast rule because apps come in so many forms.

RE: **IS_NOT_POPULATED**
Ah glad you pointed that out. I didn't know it was not considered a property.
For flexibility (or coincidence) I made the API to allow custom validators (using names) but also can be a property (can be named the same).

RE: **Memory (cleanup)**
I would love to profile our UIs and have our teams work together to address the komet platform (klworkspace) when windows close. things to think about like: _Should we have a lifecycle (cleanup method) that gets called?_ 

New in JavaFX 21 is the use of Subscribers, and Cognitive MVVM may take advantage of this facility. A concern of mine is making sure view models and change listeners get garbage collected properly. 
Just food for thought. ;-) (reduce nodes, and bindings, etc.).

besidev:
@carldea 
Just added the `GitHub Info` dialog to display some repository information. The PR is ready for a final review and testing.

besidev:
> Looks amazing! @besidev @indritbeqiri We've decided to approve and merge (along with the workaround for `.save()` (sleep).

That's great news Carl! Thanks!

PR #468, opened 2025-05-15
Feature/finished/iia 2029 edit pattern field throws exceptions

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2029

Fixes defects around patterns and the UI concept component control.  Basically there were different datatypes besides EntityProxy or EntityFacade that can get into that control and/or get muddled at save/update time.  This addresses that.

image

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
@@ -2,6 +2,7 @@ package dev.ikm.komet.kview.controls;
 
 import dev.ikm.komet.kview.controls.skin.KLComponentControlSkin;
 import dev.ikm.tinkar.common.id.PublicId;
+import dev.ikm.tinkar.entity.ConceptRecord;
 import dev.ikm.tinkar.entity.Entity;
 import dev.ikm.tinkar.terms.EntityProxy;
 import javafx.beans.property.BooleanProperty;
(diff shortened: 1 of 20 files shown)


kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
```diff
@@ -109,8 +110,20 @@ public final void setTitle(String value) {
      */
     private final ObjectProperty<EntityProxy> entityProperty = new SimpleObjectProperty<>(this, "entity", null);
     public final ObjectProperty<EntityProxy> entityProperty() { return entityProperty; }
-    public final EntityProxy getEntity() { return entityProperty.get(); }
-    public final void setEntity(EntityProxy value) { entityProperty.set(value); }
+    public final EntityProxy getEntity() {
+        if (((Object)entityProperty.get()) instanceof ConceptRecord) {
```
carldea, line 114: Do you need to down cast to an `Object`?
looks redundant.
New Java pattern matching allows variables.
```java
 if (entityProperty.get() instanceof ConceptRecord myConceptRec) {
     // do work
 } 
```

kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
```diff
@@ -109,8 +110,20 @@ public final void setTitle(String value) {
...
+        }
+        return entityProperty.get();
+    }
+    public final void setEntity(EntityProxy value) {
+        if (((Object)value) instanceof ConceptRecord) {
+            ConceptRecord conceptRecord = (ConceptRecord) (Object) value;
+            entityProperty.set(Entity.getFast(conceptRecord.nid()).toProxy());
+        }
+        entityProperty.set(value);
+    }
```
dukke, line 126: I wouldn't put this code here. Can we possibly get it into somewhere else?
IMO the control should be as much as possible decoupled from the rest of Komet.
So, as much as possible not have it use any of Komets classes, especially database objects.

kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
```diff
@@ -109,8 +110,20 @@ public final void setTitle(String value) {
...
+            entityProperty.set(Entity.getFast(conceptRecord.nid()).toProxy());
+        }
+        return entityProperty.get();
+    }
+    public final void setEntity(EntityProxy value) {
+        if (((Object)value) instanceof ConceptRecord) {
+            ConceptRecord conceptRecord = (ConceptRecord) (Object) value;
+            entityProperty.set(Entity.getFast(conceptRecord.nid()).toProxy());
+        }
+        entityProperty.set(value);
```
carldea, line 125: This looks odd to me.
Should the if condition return? It's setting `entityProperty.set()` twice.
We should ask for a second opinion from the data team.

dukke (changes requested):
Nice work fixing this however I find having database classes inside the custom controls not ideal. I think as a design goal it's better to keep the custom controls as much decoupled from the rest of Komet as possible, especially database objects.

kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
dholubek: the controls already have EntityFacade typed in them.

kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
dholubek: there are times when it is one data type and times when it is another... ie ConceptRecord and EntityFacade.

PR #469, opened 2025-05-16
Feature/finished/iia 1993 negative number fix

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1993

The defect exists when you have a single digit negative number and trying to hit the backspace key from the right of the number will be stuck and not allow the user to delete it.  This fix allows the user to continue that delete operation.  It isn't perfect but it is an improvement


carldea (approved):
Looks good to me.
We'll need to revisit this for an improved solution.

PR #470, opened 2025-05-16
feature/finished/IIA-1992-fix-concurrent-modification-exception-copy-list

jdsmithsos:
This is a quick fix for the ConcurrentModificationException.
Carl could reproduce on Mac, which behaves differently than Windows with concurrent access to the ArrayList during iteration.
Changed to not use the iterator directly to using a copy of the listener list instead.


carldea (approved):
Looks good to me!
We can revisit if time permits.

PR #473, opened 2025-05-18
Feature/finished/iia 1982 c set type ahead duplicate component

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-1982

This fixes a bug in typeahead where the user was able to add a duplicate concept in a CSet while using typeahead. 
I think it's on the frontier between being a must have for may's demo and not being needed for may's demo.
The thing is the bug breaks the CSet feature of not allowing for duplicates however whoever is demoing can demo around that bug.

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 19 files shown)


carldea (approved):
Looks good to me.

carldea (approved):
Looks good to me.

PR #478, opened 2025-05-19
Feature/finished/iia 2027 resolve create semantic and fixes

carldea:
This branch contains.

1. fix to exception when creating a semantic
2. fix to use window topic instead of journal topic
3. support many dnd sources such as original concept nav


kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentControlSkin.java
```diff
@@ -216,34 +232,141 @@ private void setupDragNDrop() {
...
                 try {
+                    int nid = extractNid(event);
                     LOG.info("publicId: {}", dragboard.getString());
-                    if (event.getGestureSource() instanceof Node source &&
-                            source.getUserData() instanceof DragAndDropInfo dropInfo &&
-                            dropInfo.publicId() != null
-                    ) { // TODO: should this be needed? shouldn't we get PublicId from dragboard content?
-                        int nid = EntityService.get().nidForPublicId(dropInfo.publicId());
-                        EntityProxy entity = EntityProxy.make(nid);
+                    if (nid != Integer.MIN_VALUE) {  //
```
dukke, line 237: We stopped using Integer.MIN_VALUE to denote an empty component control. 
There's a method called isEmpty() that can be used, here would be `getSkinable().isEmpty()`

dukke (commented):
@carldea Just had a quick look as it's a bit late here :) Added 1 comment from that very brief look.

application/pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.51.0-SNAPSHOT</version>
+        <version>1.51.0-IIA-2027-resolve-create-semantic-and-fixes-SNAPSHOT</version>
     </parent>
```
dholubek, line 8: @carldea looks like the feature-finish didn't finish... probably want to find/replace and push this manually.

carldea: Thanks, I did a find/replace.

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentControlSkin.java
carldea: Please examine the method extractNid(event). if not found it will use this reserved integer to denote not found.
All this code should be refactored to keep much of this database specific code out of the skin. But this was much needed as the user will want to drag components from different ui controls as before.

framework/src/main/java/dev/ikm/komet/framework/view/ViewMenuTask.java
```diff
@@ -560,9 +560,11 @@ private static void addChangeItemsForView(ViewCalculator viewCalculator, List<Me
             Platform.runLater(() -> {
                 MenuItem sourceMenu = (MenuItem) event.getSource();
                 Menu parentMenu = sourceMenu.getParentMenu();
-                parentMenu.getItems().clear();
-                TinkExecutor.threadPool().execute(TaskWrapper.make(new ViewMenuTask(viewCalculator, observableView),
-                        (List<MenuItem> result) -> parentMenu.getItems().addAll(result)));
+                if (parentMenu != null) {
+                    parentMenu.getItems().clear();
+                    TinkExecutor.threadPool().execute(TaskWrapper.make(new ViewMenuTask(viewCalculator, observableView),
+                            (List<MenuItem> result) -> parentMenu.getItems().addAll(result)));
```
dukke, line 566: Is this menu already attached to a Scene? If so we'd want to use Platform.runLater to execute the code that changes the UI in the UI thread

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
```diff
@@ -346,10 +349,24 @@ public void submit(ActionEvent actionEvent) {
...
            SemanticVersionRecord version = Entity.getVersionFast(semantic.nid(), stamp.nid());
            Transaction.forVersion(version).ifPresentOrElse(transaction -> {
-               commitTransactionTask(transaction);
-               // EventBus implementation changes to refresh the details area if commit successful
-               EvtBusFactory.getDefaultEvtBus().publish(genEditingViewModel.getPropertyValue(CURRENT_JOURNAL_WINDOW_TOPIC),
-                       new GenEditingEvent(actionEvent.getSource(), PUBLISH, list, semantic.nid()));
+
+                   try {
+//                       EntityService.get().endLoadPhase();
+                       createSemanticVersionTransactionTask(transaction).get(); // get() will block until transaction is finished (will refresh view calculator caches)
```
dukke, line 355: We are inside the UI thread and we are blocked until get() returns, this will block the UI thread until the task is complete, and the UI will become unresponsive. 
We could spawn a new thread. Inside that thread we run this code and after we have the results call all the code that updates the UI inside Platfrom.runLater

dukke (changes requested):
Added some more comments.

I appreciate the commented out text saying we would get the database objects out of this control, I hope we have time to do it :)

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
```diff
@@ -479,11 +489,11 @@ private void setupReferenceComponentUI() {
                 ObjectProperty<EntityFacade> newRefComponentProp = genEditingViewModel.getProperty(REF_COMPONENT);
                 updateRefComponentInfo.accept(newRefComponentProp.get());
                 //Enable disable pencil icons
-                editFieldsButton.disableProperty().bind(newRefComponentProp.isNull());
-                addReferenceButton.disableProperty().bind(newRefComponentProp.isNotNull());
+                editFieldsButton.setDisable(newRefComponentProp.isNull().get());
+                addReferenceButton.setDisable(newRefComponentProp.isNotNull().get());
```
dukke, line 493: if we use binding above we won't need to update the disable property of the buttons again here

dukke (changes requested):

kview/src/main/java/dev/ikm/komet/kview/controls/skin/KLComponentControlSkin.java
```diff
@@ -216,34 +232,141 @@ private void setupDragNDrop() {
...
+        // Detect classic search to drag and drop components (concepts and semantics)
+        boolean classicSearchDnD = isFromClassicSearch(dragEvent);
+
+        // Detect new concept navigator to drag and drop concepts
+        boolean nextGenConceptNavigatorDnD = isFromNexGenConceptNav(dragEvent);
+
+        // Existing checks for next gen search navigator items to drag.
+        boolean nextGenSearchDnD =  isADragAndDropInfo(dragEvent);
+
+        return classicConceptNavigatorDnD || classicSearchDnD || nextGenConceptNavigatorDnD || nextGenSearchDnD;
```
dholubek, line 281: @carldea this is fixing what we can drop into a component, which is good, but it has a broad scope because the component is used by the Semantic window and the Pattern window in several places.  So the permutations of testing all this is broad with classic Komet and next gen search against those two window types and all the places inside them that use the component also during edit and create modes... are we sure we want to introduce that scope this late?  Do we want to just stick to the transaction related fix on the Semantic?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
swaroopsalvi: This was discussed and Carl specifically wanted to not have bind here. He has done this intentionally.

swaroopsalvi (approved):
Pros:
   1. Fixing create semantic issue (Exception thrown during creation fixed).
   2. Support for classic komet navigator drag and dorp and search -> Unit tested and ready.
   3. Integration tinkar-core 1.103.0 -> performance related fix in tinkar-core.
   4. Unblocks current sprint stories for updateVersion and semantic.

Cons:
   1. This is a last min change and will need some additional testing.
   2. It will freeze the UI while the save is in progress. (Side effect of the solution done.) -> Can be fixed.
   3. One of JavaFx rules to have Platform.runLater() to be used to update the UI code.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
```diff
@@ -223,11 +224,11 @@ public GenEditingDetailsController() {
     @FXML
     private void initialize() {
 
-        EntityFacade refComponent = genEditingViewModel.getPropertyValue(REF_COMPONENT);
+        ObjectProperty<EntityFacade> refComponent = genEditingViewModel.getObjectProperty(REF_COMPONENT);
         //Enable edit fields button if refComponent is NOT null else disable it.
-        editFieldsButton.setDisable(refComponent == null);
+        editFieldsButton.setDisable(refComponent.isNull().get());
```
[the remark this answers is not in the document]
carldea: Good call! We should create another ticket since this is merged already.

framework/src/main/java/dev/ikm/komet/framework/view/ViewMenuTask.java
carldea: I am late getting these messages. It's too late in the game. But next time small stuff like this feel free to fix as a shared branch.
I agree with @dholubek  this is read only for the demo and not to be invoked (trigger actions). But I believe the code was really lifted from classic Komet. Ideally when using a thread pool it provides activity information in the Activities tab for long running loads (building many menu options). 

I think @dukke  is correct that the parent (context menu) is attached to a Scene, and when adding to it (parentmenu) should be done on the JavaFX thread.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
carldea: Pedro is correct in that we should never block like this, but while fixing this issue we needed it to block in order to know diagnose why/when a transaction is finished creating a semantic. Swaroop was half/correct (from yesterday), but because of the fix in tinkar core we'd need to put the async code back in (surrounding the code above). 

Let me explain.
Using a small dataset, the response time is negligible(fast), however if you use the snomed data, the creation code takes a very long time. Yesterday we found out the issue was code in the database code was asynchronous and we couldn't predict when the transaction was actually complete and the left side would query too soon and retrieve data from a stale cache.

Once we fixed that issue hence tinkar-core 1.103.0. I forgot to add the Completable futures code back in to not freeze the GUI. :(

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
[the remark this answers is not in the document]
dukke: @carldea yes, a ticket is a good idea! Do you want me to create it?
 There are a bunch of places perhaps we should clean up.. including my code (for instance KLComponentSet control is still in the repo but we don't need it anymore).
Perhaps we could take some time after the demo to reduce some of our technical debt?

framework/src/main/java/dev/ikm/komet/framework/view/ViewMenuTask.java
dukke: @carldea sounds good. Do you want me to add this to the ticket we mentioned?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/SemanticFieldsController.java
dukke: @carldea Yeah now that you mentioned it, I do remember that in the code you posted on Teams in our group chat you were indeed creating a new thread and calling this code inside that thread.
I did a small test where I simulated a long running call to the database (using thread sleep and having the code be the same as in this PR) and indeed the UI completely froze.
I have a quick fix idea for this if you'd like me to share it. 
We could change the `createSemanticVersionTransactionTask` method to have an additional parameter that would be a callback that is called when the transactions is finished. Something like this:
``
private Future<Void> createSemanticVersionTransactionTask(Transaction transaction, Runnable runAfterTransaction) {
        CommitTransactionTask commitTransactionTask = new CommitTransactionTask(transaction);
        return TinkExecutor.threadPool().submit(new Callable<Void>() {
            @Override
            public Void call() throws Exception {
                commitTransactionTask.call();
                runAfterTransaction.run();
                return null;
            }
        });
    }
``
Notice the runAfterTransaction runnable callback.

PR #479, opened 2025-05-20
feature/finished/IIA-1588-add-tooltip-on-pattern-browser

jdsmithsos:
Jira ticket:  

https://ikmdev.atlassian.net/browse/IIA-1588

Added a ToolTip to the semantic label in the Pattern Browser.

Before change:

No ToolTip.

image

After change:

With ToolTip.

image


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/navigation/PatternNavEntryController.java
```diff
@@ -255,6 +246,7 @@ protected void updateItem(Object item, boolean empty) {
 
                         setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
                         label.setText(entityDescriptionText);
+                        Tooltip.install(label, new Tooltip(entityDescriptionText));
```
dukke, line 249: it's best to create this tooltip outside of the updateItem method. This updateItem method gets called everytime a cell gets updates with a new value. So you'll be creating lots of Tooltip objects this way.

dukke (changes requested):
Left one request for change. Thanks!

carldea (approved):
Looks good to me!

PR #481, opened 2025-05-20
IIA-1854 import button implementation on Landing Page

bhharsh13:
Implemented the Import button functionality on the Landing Page.
Validated on local  and looks good. PFA for reference.
Please help to review the PR.
import button in Landing page

Selected Dataset to import

### DIFF
--- a/application/src/main/java/dev/ikm/komet/app/App.java
+++ b/application/src/main/java/dev/ikm/komet/app/App.java
@@ -24,6 +24,7 @@ import dev.ikm.komet.framework.KometNodeFactory;
 import dev.ikm.komet.framework.ScreenInfo;
 import dev.ikm.komet.framework.activity.ActivityStreamOption;
 import dev.ikm.komet.framework.activity.ActivityStreams;
+import dev.ikm.komet.framework.events.Evt;
 import dev.ikm.komet.framework.events.EvtBus;
 import dev.ikm.komet.framework.events.EvtBusFactory;
 import dev.ikm.komet.framework.events.Subscriber;
@@ -144,6 +145,7 @@ import static dev.ikm.komet.app.util.CssFile.KVIEW_CSS;
 import static dev.ikm.komet.app.util.CssUtils.addStylesheets;
 import static dev.ikm.komet.framework.KometNode.PreferenceKey.CURRENT_JOURNAL_WINDOW_TOPIC;
 import static dev.ikm.komet.framework.KometNodeFactory.KOMET_NODES;
+import static dev.ikm.komet.framework.events.FrameworkTopics.IMPORT_TOPIC;
 import static dev.ikm.komet.framework.window.WindowSettings.Keys.CENTER_TAB_PREFERENCES;
 import static dev.ikm.komet.framework.window.WindowSettings.Keys.LEFT_TAB_PREFERENCES;
 import static dev.ikm.komet.framework.window.WindowSettings.Keys.RIGHT_TAB_PREFERENCES;
@@ -442,6 +444,15 @@ public class App extends Application {
             stage.show();
             state.set(AppState.SELECT_DATA_SOURCE);
             state.addListener(this::appStateChangeListener);
+
+            Subscriber<Evt> importSubscriber = _ -> {
+                /**
+                 * Handles the import button event by initiating the import process if the ViewModel is valid.
+                 * */
+                openImport();
+            };
+            kViewEventBus.subscribe(IMPORT_TOPIC, Evt.class, importSubscriber);
+
         } catch (IOException e) {
             LOG.error(e.getLocalizedMessage(), e);
             Platform.exit();
--- a/application/src/main/java/dev/ikm/komet/app/App.java
+++ b/application/src/main/java/dev/ikm/komet/app/App.java
@@ -445,8 +445,8 @@ public class App extends Application {
             state.set(AppState.SELECT_DATA_SOURCE);
             state.addListener(this::appStateChangeListener);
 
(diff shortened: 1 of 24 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/landingpage/LandingPageController.java
```diff
@@ -410,4 +421,27 @@ void createNewJournalViewFromCard(Event event) {
     public Label getWelcomeTitleLabel() {
         return welcomeTitleLabel;
     }
+
+    /**
+     * Handles the import button event by initiating the import process if the ViewModel is valid.
+     * */
+    @FXML
+    private void openImport() {
+        KometPreferences appPreferences = KometPreferencesImpl.getConfigurationRootPreferences();
```
dholubek, line 430: @bhharsh13 this openImport() method will live in 3 places with this change kview/LandingPageController, application/App and application/WebApp... I'm wondering if there is a good place to refactor so that it is in one place but I don't want to create a circular dependency because: application > kview > framework... so framework can't depend on kview and it doesn't make a ton of sense to put it in application... we could of course leave it as is...

thoughts? @carldea

dholubek (commented):
I have an open question about whether this should be refactored... rule of 3 (possibly) = when we encounter it 3 times then it might be a good time to make it common.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/landingpage/LandingPageController.java
carldea: Yes, very good point.
Create an event to be published. The App and Web App will need to a listener to respond. 
Both classes own the method that invokes the import window. 
Both classes have access to the Event in kview, and would be able to respond when event is published.

The landing page's import button, will publish the event. Since there is only one Landing page, it would use a unique topic, and any other information if needed. But I don't think so.

bhharsh13: Created an event to listen to IMPORT_TOPIC  for App and Web App to listen. Validated changes on local and looks good.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/landingpage/LandingPageController.java
```diff
@@ -410,4 +413,12 @@ void createNewJournalViewFromCard(Event event) {
     public Label getWelcomeTitleLabel() {
         return welcomeTitleLabel;
     }
+
+    /**
+     * Handles the import button event by initiating the import process if the ViewModel is valid.
+     * */
```
carldea, line 419: Can you fix the end of the javadoc comment? Also, the sentence should read as:
"_Handles the import button press to pop up the import dialog window._"
Add javadoc annotations for the `event` parameter.

carldea:
> I have an open question about whether this should be refactored... rule of 3 (possibly) = when we encounter it 3 times then it might be a good time to make it common.

good call. I think it's a rule of 3. In WebApp the dialog needs a owner stage window and App.java just takes the primary stage window. So yea, It is a very good candidate for a refactor. after refactoring (in another ticket) each place needing the popup import can simply emit the event on the `IMPORT_TOPIC` (in 3-4 locations).

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/landingpage/LandingPageController.java
bhharsh13: Modified the javadoc comment

application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -442,6 +444,15 @@ public void start(Stage stage) {
             stage.show();
             state.set(AppState.SELECT_DATA_SOURCE);
             state.addListener(this::appStateChangeListener);
+
+            Subscriber<Evt> importSubscriber = _ -> {
+                /**
+                 * Handles the import button event by initiating the import process if the ViewModel is valid.
```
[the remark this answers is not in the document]
bhharsh13: Modified to use inline comment

application/src/main/java/dev/ikm/komet/app/App.java
```diff
@@ -442,6 +444,13 @@ public void start(Stage stage) {
             stage.show();
             state.set(AppState.SELECT_DATA_SOURCE);
             state.addListener(this::appStateChangeListener);
+
+            Subscriber<Evt> importSubscriber = _ -> {
+                //Pops up the import dialog window on any events received on the IMPORT_TOPIC
```
carldea, line 449: Can we move it above the declaration? before line 448.

application/src/main/java/dev/ikm/komet/app/WebApp.java
```diff
@@ -377,6 +379,12 @@ public void init() throws Exception {
 
         // Subscribe the subscriber to the USER_TOPIC
         kViewEventBus.subscribe(USER_TOPIC, SignInUserEvent.class, signInUserEventSubscriber);
+
+        Subscriber<Evt> importSubscriber = _ -> {
+            //Pops up the import dialog window on any events received on the IMPORT_TOPIC
```
carldea, line 384: can we move this above line 383?

application/src/main/java/dev/ikm/komet/app/App.java
bhharsh13: moved the comment as suggested

carldea (approved):
Things look good!

PR #483, opened 2025-05-22
Feature/finished/iia 1398 semantic creation close properties

jdsmithsos:
Jira ticket:

https://ikmdev.atlassian.net/browse/IIA-1398

Summary of changes:

- Specific changes for the mode were not required, e.g. CREATE vs EDIT.
    - The text will be verified when I demo to Jieun for the different scenarios of when the Confirmation pane is displayed
- Modified ClosePropertiesController to follow same pattern as ConfirmationController, where the controller can be reused to display different text
- Modified pattern/PropertiesController to call the ConfirmationController show methods to set the text to be displayed

Initial window for creation of a Semantic:

image

Selecting reference component:

image

image

Entering String property value:

image

Before changes, pressing Submit::

Title text is "No Selection Made"

image

After changes, pressing Submit:

Title text is now "Semantic Details Added"

image


carldea:
A bit concerned with 3 String instance variables in the controller to display a simple confirmation message text.

**Suggestions:**
You may want to use a view model (`SimpleViewModel`) or Not (it could be overkill). However, A view model can have a property confirmation text. The ui's labels would be bound to the property. You can communicate using updateModel(). Also, make use of the event bus, as each window's submit or confirm, can emit an event maybe the text to show.

**MVVM think**
Controllers (mvvm pattern) are considered views. Presentation logic would be inside view models. Once you get the hang of it'll be easy. UI developers tend to have controllers that are bloated with complex database/a mix of ui code. 
ViewModels should never have UI things inside such as labels, buttons, etc.

**Check out the wiki:**
https://github.com/carldea/cognitive/wiki

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
```diff
@@ -109,15 +103,16 @@ private void setupShowingPanelHandlers() {
...
-
         closePropsPane = closePropsJfxNode.node();
         closePropertiesController = closePropsJfxNode.controller();
+
         genEditingEventSubscriber = evt -> {
             LOG.info("Publish event type: " + evt.getEventType());
-            contentBorderPane.setCenter(closePropsJfxNode.node());
+
+            // "Semantic Details Added" is displayed when form values are Submitted
+            closePropertiesController.showSemanticDetailsAdded();
```
swaroopsalvi, line 114: I will recommend to use the following logic:
            if (genEditingViewModel.getPropertyValue(MODE) == CREATE) {
                closePropertiesController.setHeadingText("Semantic Details Added");
            }else {
                closePropertiesController.setHeadingText("Semantic Details Changed");
            }

and no need to modify the ClosePropertiesController. Will be less code foot print and use of MVVM as well.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/ClosePropertiesController.java
```diff
@@ -36,13 +36,15 @@ public class ClosePropertiesController {
...
 
-
     @FXML
     private Label subtextLine1;
 
     @FXML
     private Label subtextLine2;
 
+    private String headingTextDefault;
+    private String subtextLine1Default;
```
swaroopsalvi, line 46: Can remove this variables. Since the default subtextLine1 and subtextLine2 would always be the same.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/ClosePropertiesController.java
```diff
@@ -53,6 +55,13 @@ private void closeProperties(ActionEvent event) {
         EvtBusFactory.getDefaultEvtBus().publish(genEditingViewModel.getPropertyValue(WINDOW_TOPIC), new PropertyPanelEvent(event.getSource(), CLOSE_PANEL));
     }
 
+    @FXML
+    public void initialize() {
+        headingTextDefault = headingText.getText();
```
swaroopsalvi, line 60: No need to save the default header value we can load them from Properties controller. See the code below.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/ClosePropertiesController.java
```diff
@@ -76,4 +85,17 @@ public void setSubtextLine1(String text) {
     public void setSubtextLine2(String text) {
         subtextLine2.setText(text);
     }
+
+    public void showNoSelectionMadeToEditSemanticElement() {
+        headingText.setText("No Selection Made");
```
swaroopsalvi, line 90: Can set this directly from Properties Controller.

swaroopsalvi (changes requested):
The only change that I think should be done is in PropertiesController.
by using the MVVM.
  The code will be: 
  if (genEditingViewModel.getPropertyValue(MODE) == CREATE) {
                closePropertiesController.setHeadingText("Semantic Details Added");
            }else {
                closePropertiesController.setHeadingText("Semantic Details Changed");
            }

carldea:
> The only change that I think should be done is in PropertiesController. by using the MVVM. The code will be: if (genEditingViewModel.getPropertyValue(MODE) == CREATE) { closePropertiesController.setHeadingText("Semantic Details Added"); }else { closePropertiesController.setHeadingText("Semantic Details Changed"); }

good suggestion to reuse the existing "shared" view model across controllers. But **not have methods** on the controller. 
And to add a property on the `genEditingViewModel.setPropertyValue(CONFIRMATION_MESSAGE_TEXT, "text goes hear");`

The proper way would be in the `ClosePropertiesController.initialize()` method you would have a `Label confirmationTextMessage` (probably in FXML which means you need an fx:id). Assuming you have a reference to the label.
```java
// instance variables
@FXML
Label confirmationTextMessage;

// inside the initialize() method
confirmationTextMessage.textProperty().bind(genEditingViewModel.setProperty(CONFIRMATION_MESSAGE_TEXT));
```
**Changing confirmation text:**
Since every controller has a reference to the `genEditingViewModel` (they/ a controller) can update the property after a submit.
```java
genEditingViewModel.setPropertyValue(CONFIRMATION_MESSAGE_TEXT, "Semantic Details Added");
```

No need to use the event bus since our last sprint **all controllers for GenEditing Window** should be sharing the `genEditingViewModel` view model. Because of this the UI Form state is maintained for a window and all child controllers are also referencing properties (presentation data/logic).

PR #490, opened 2025-05-28
Feature/finished/iia 1932 semantic for ref component

dholubek:
https://ikmdev.atlassian.net/browse/IIA-1932

able to use semantics on reference component, component, cList and cSet and to show the name properly since entityProxy.description() doesn't work for semantics.  Tested creating and editing semantics against patterns, classic concepts, next gen search concepts and type ahead in those component fields and it is working.

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
@@ -15,9 +15,11 @@ import javafx.beans.property.StringProperty;
 import javafx.collections.MapChangeListener;
 import javafx.event.ActionEvent;
 import javafx.event.EventHandler;
-import javafx.scene.Node;
 import javafx.scene.control.Control;
+import javafx.scene.control.ListCell;
+import javafx.scene.control.ListView;
 import javafx.scene.control.Skin;
+import javafx.util.Callback;
 import javafx.util.StringConverter;
 
 import java.util.List;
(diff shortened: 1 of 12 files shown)


kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
```diff
@@ -127,6 +126,16 @@ public final void setEntity(EntityProxy value) {
     public final Function<String, List<EntityProxy>> getCompleter() { return completer.get(); }
     public final ObjectProperty<Function<String, List<EntityProxy>>> completerProperty() { return completer; }
 
+    // -- function to render the component's name and avoid entity.description()
+    private final ObjectProperty<Function<EntityProxy, String>> componentNameRenderer = new SimpleObjectProperty<>();
+    public final Function<EntityProxy, String> getComponentNameRenderer() { return componentNameRenderer.get(); }
+    public final void setComponentNameRenderer(Function<EntityProxy, String> nameHandler) {
+        componentNameRenderer.set(nameHandler);
+    }
+    public final ObjectProperty<Function<EntityProxy, String>> getComponentNameRendererProperty() {
```
dukke, line 135: in javafx convention this should be named componentNameRendererProperty()..

So, the javafx convention for properties is:
Getters - get<name> or is<name> (in case of boolean)
Setter - set<name>
Property getter - <name>Property

kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentListControl.java
```diff
@@ -98,6 +98,16 @@ public final void setTitle(String value) {
     public final Function<EntityProxy, Node> getSuggestionsNodeFactory() { return suggestionsNodeFactory.get(); }
     public final ObjectProperty<Function<EntityProxy, Node>> suggestionsNodeFactoryProperty() { return suggestionsNodeFactory; }
 
+    // -- function to render the component's name and avoid entity.description()
+    private final ObjectProperty<Function<EntityProxy, String>> componentNameRenderer = new SimpleObjectProperty<>();
+    public final void setComponentNameRenderer(Function<EntityProxy, String> nameHandler) {
+        componentNameRenderer.set(nameHandler);
+    }
+    public final Function<EntityProxy, String> getComponentNameRenderer() { return componentNameRenderer.get(); }
+    public final ObjectProperty<Function<EntityProxy, String>> getComponentNameRendererProperty() {
```
dukke, line 107: same here as the comment above about the correct name for property getters

dukke (changes requested):
Looks good to me!
Only had one comment which is about the naming convention (on 2 places).
Thanks

kview/src/main/java/dev/ikm/komet/kview/controls/KLComponentControl.java
dholubek: fixed

carldea (approved):
Looks good to me!
nice work.

PR #492, opened 2025-05-28
NOT TO EVER BE MERGED - IIA-2079 added remote debugging capability on port 6006

jdsmithsos:
added an option for remote JPDA debugging on port 6006


dholubek:
@jdsmithsos we have a debug komet image that is build with maven today.  I'm not sure if this already solves the problem that you are looking to solve... in windows you run the .bat file and then from IntelliJ you do Run > Attach to Process

image

carldea:
This has been pending for a while. I closed it.

PR #494, opened 2025-05-28
Feature/finished/iia 2067 type ahead cell factory

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2067

This PR should further increase the perceived speed of search and also the amount of memory used. (I say perceived speed of search because it increases the speed of typeahead while you're typing and not of the actual search that happens after you hit enter)
It changes the mechanist TypeAhead uses internally to show the suggestions.


carldea (approved):
Looks good to me!
Now it is easy to configure cell factories.

PR #495, opened 2025-05-28
IIA-2043- Fixed Concept Definition Description type to be not shown in Other Names Descriptions

bhharsh13:
https://ikmdev.atlassian.net/browse/IIA-2043

Fixed Concept Definition Description type to be not shown in Other Names Descriptions. 
Validated on local and changes looks good.

image


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
```diff
@@ -1009,17 +1009,16 @@ public void updateConceptDescription() {
             PatternEntity<PatternEntityVersion> patternEntity = semanticEntityVersion.pattern();
             PatternEntityVersion patternEntityVersion = viewCalculator.latest(patternEntity).get();
 
-            boolean isFQN = semanticEntityVersion
-                    .fieldValues()
-                    .stream()
-                    .anyMatch( fieldValue ->
-                            (fieldValue instanceof ConceptFacade facade) &&
-                                    facade.nid() == FULLY_QUALIFIED_NAME_DESCRIPTION_TYPE.nid());
+            int descFieldType = getFieldIndexByMeaning(semanticEntityVersion, TinkarTerm.DESCRIPTION_TYPE);
```
carldea, line 1012: Is there a way to move this statement outside of the forEach loop?

dholubek: @carldea I was taking a look with @bhharsh13 

1) Do you want a .filter() before the .forEach() and if yes, 
1a)  How do we utilize the DESCRIPTION_PATTERN to do the filtering?  did you mean to write DESCRIPTION_TYPE?  I'm not sure how we do the filtering without this index look up...???
2) I was playing around with setting up a filter() and a map() but at a minimum the body of the forEach still needs both the semanticEntityVersion and the list of fieldValues to pass onto the helper methods to build the UI

at first I thought you just wanted a filter something like this but I'm guessing you won't like this:

`
descriptionSemanticsMap.keySet().stream()
                .filter(semanticEntityVersion -> getFieldIndexByMeaning(semanticEntityVersion, 
TinkarTerm.DESCRIPTION_TYPE) != 1)
                .forEach((semanticEntityVersion) -> {
`

we still have to do the index lookup inside the for loop even with that filter to grab the correct field value, so I'm guessing that is not what you are looking for... I played around with a map() function as well so that it would return a List.of(semanticEntityVersion, conceptFacade /* to avoid the index look up in the for loop */, fieldValues)
since we still need the semanticEntityVersion adn the fieldValues... but that code looked extremely messy...

bhharsh13: Moved the DescriptionTypeIndex out of ForEach Loop

PR #497, opened 2025-05-29
Feature/Finished/IIA-2066: We should store empty components using Integer.MIN_VALUE, IIA-1910, IIA-1915

dukke:
Fixes: 
1. https://ikmdev.atlassian.net/browse/IIA-2066
2. https://ikmdev.atlassian.net/browse/IIA-1910
3. https://ikmdev.atlassian.net/browse/IIA-1915

Empty components are stored using Integer.MIN_VALUE as it is a reserved value in the database. That would also start to be the default value for component field.

Side note and for future reference: While using Integer.MIN_VALUE for empty component fields, we cannot call toString on the observable when it is storing that value as Tinker will throw an exception in that case.

### DIFF
--- a/framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
+++ b/framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
@@ -104,11 +104,10 @@ public final class ObservableSemanticVersion
                  && (observableField.fieldProperty.getValue().value() != null)) {
                     // Check if the previous value is different from the changed value.
                     // This check is required for C-List C-Set
-                    if (observableField.value() instanceof IntIdSet || observableField.value() instanceof IntIdList) {
-                        IntIdCollection valueList = (IntIdCollection) observableField.value();
-                        IntIdCollection fileValueList = (IntIdCollection) observableField.fieldProperty.getValue().value();
+                    if (observableField.value() instanceof IntIdCollection valueList) {
+                        IntIdCollection fieldValueList = (IntIdCollection) observableField.fieldProperty.getValue().value();
 
-                        if (!Arrays.equals(valueList.toArray(), fileValueList.toArray())) {
+                        if (!Arrays.equals(valueList.toArray(), fieldValueList.toArray())) {
                             // Creating uncommitted version records. e.g., (c)hello, (u)hello1, (u)hello12, (u)hello123
                             autoSaveSematicVersion(observableField.value(), index);
                         }
--- a/framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
+++ b/framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
@@ -15,6 +15,9 @@
  */
 package dev.ikm.komet.framework.observable;
 
+import dev.ikm.tinkar.common.id.IntIdCollection;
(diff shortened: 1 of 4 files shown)


framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
```diff
@@ -98,13 +101,21 @@ public ImmutableList<ObservableField> fields(PatternEntityVersion patternVersion
...
                 // ObservableSemanticVersion owns the versionProperty<SemanticVersion>
                 if (observableField.value() != null // Create a version only when new value is not null.
-                 && (observableField.fieldProperty.getValue().value() != null &&
-                 // Check if the previous value is different from the changed value.
-                 // This check is required for C-List C-Set
-                 !Objects.equals(observableField.value().toString(), observableField.fieldProperty.getValue().value().toString()))
-                ) {
-                    // Creating uncommitted version records. e.g., (c)hello, (u)hello1, (u)hello12, (u)hello123
-                    autoSaveSematicVersion(observableField.value(), index);
+                 && (observableField.fieldProperty.getValue().value() != null)) {
```
swaroopsalvi, line 104: This should not be part of this PR, we need to create a new PR for this code change and discuss this before we implement this change.

swaroopsalvi (changes requested):
I have one comment for the code in ObservableSemanticVersion

framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
dukke: @swaroopsalvi the problem is that calling toString on a EntityProxy that has Integer.MIN_VALUE (which this PR is using) blows things up. It throws an expcetion in the backend that's why I changed this here

swaroopsalvi: Got it. Have one more comment in the below logic.

framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
```diff
@@ -98,13 +101,21 @@ public ImmutableList<ObservableField> fields(PatternEntityVersion patternVersion
...
+                    // Check if the previous value is different from the changed value.
+                    // This check is required for C-List C-Set
+                    if (observableField.value() instanceof IntIdSet || observableField.value() instanceof IntIdList) {
```
swaroopsalvi, line 107: Can we have this if condition as:
 if (observableField.value() instanceof IntIdCollection intIdCollection){
   // change the inside logic appropriately?
}

swaroopsalvi (changes requested):

framework/src/main/java/dev/ikm/komet/framework/observable/ObservableSemanticVersion.java
dukke: @swaroopsalvi yeah, good idea!

dukke: @swaroopsalvi  done!

swaroopsalvi (approved):
Approved.
Thanks.

dukke:
Closing this PR as was discussed with @carldea

PR #503, opened 2025-06-02
Feature/finished/iia 1547 refactor code to use helper class for accessing concept descendants

swaroopsalvi:


carldea (approved):
@swaroopsalvi Excellent work!
So glad this is getting cleaned up and based on the database.
Looks good to me!

PR #504, opened 2025-06-02
Feature/finished/iia 1588 much simpler pattern name tooltip

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-1588

Added tooltip to the Pattern Browser pattern name label.

Before change:

No tooltip.

image

After change:

Tooltip.

image

The Semantic name label tooltip was added in a previous PR.

image


carldea (approved):
Looks good to me!
Thank you @jdsmithsos

PR #509, opened 2025-06-03
Feature/finished/iia 2083 refactor closeproperties confirmation with mvvm

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2083

Summary of changes:

- Combined the close-properties.fxml and confirmation-pane.fxml, and their controllers, into a single fxml and controller class
    - confirmation-pane-common.fxml
    - ConfirmationPaneCommonController
- Created a new view model class ConfirmationPaneCommonViewModel that contains
    - title SimpleStringProperty
    - message SimpleStringProperty
    - notificiationTopic property
    - notificationEvent property
    - Enum for the various title/messages that can be displayed
    - sendNotification() method that sends the notification on the event bus to the specified notificationTopic using the notificiationEvent

Before change:

There were separate fxml and controllers for the Pattern and Semantic property editors.

In Pattern, adding Description.

image

In Pattern, after pressing Done button.

image

In Pattern, after pressing CLOSE PROPERTIES PANEL button.

image

In Semantic, editing properties.

image

In Semantic, after pressing Submit button.

image

In Semantic, after pressing CLOSE PROPERTIES PANEL button.

image

After changes:

There is now a common confirmation pane with fxml, controller, and view model.

In Pattern, adding Description.  This is the exact same as before changes.

image

In Pattern, after pressing Done button.  This uses the new common confirmation.

image

In Pattern, after pressing CLOSE PROPERTIES PANEL button.  This shows that the common controller successfully sends the event to close the properties pane on the Pattern window.

image

In Semantic, editing properties.  This is the same as before changes.

image

In Semantic, after pressing Submit button.  This uses the new common confirmation.

image

In Semantic, after pressing CLOSE PROPERTIES PANEL button.  This shows that the common controller successfully sends the event to close the properties pane on the Pattern window.

image

I didn't show it, but the Toast message also displays as it did before the change.


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/confirmation/ConfirmationPaneCommonController.java
```diff
@@ -0,0 +1,48 @@
...
+    private Label confirmationTitle;
+
+    @FXML
+    private Label confirmationMessage;
+
+    /**
+     * View model for display text and notification topic and event
+     */
+    @InjectViewModel
+    private ConfirmationPaneCommonViewModel viewModel;
```
carldea, line 31: Can we use the `SimpleViewModel` class?

The caller using `Config` object and `FXMLMvvmLoader.make()` has the ability to pass a view model or one supplied by the framework (instantiated). 
```java
   // Caller supplied view model
   ViewModel myViewModel = new SimpleViewModel();
   myViewModel.addProperty(CONF_TITLE, "");
   myViewModel.addProperty(CONF_MESSAGE, "");
   config.addNamedViewModel(new NamedVm("viewModel", myViewModel)));
  
  // Allow the framework to supply view model for the caller (code that creates the view/controller/fxml).
  Optional<ViewModel> vmOptional = jfxNode.getViewModel("viewModel");
  // you can add properties here or the preferred way is inside of the initialize() method.
  vmOptional.ifPresent( viewModel -> {
       viewModel.addProperty(CONF_TITLE, "");
      viewModel.addProperty(CONF_MESSAGE, "");
 });
 
 // Since the controller's initialize() method is called once the addProperty() method is also called once.
 // addProperty method defines and creates properties in order to be allowed to call setPropertyValue() or getPropertyValue() methods.    
  
```

**Reason:** when there isn't complex logic or advanced scenarios with the UI form use the framework's concrete `SimpleViewModel`. When a form calls for validation logic you can use the `ValidationViewModel`. Creating a custom view model is good when there are other scenarios such as different views pointing to a view model e.g. Semantic editing window with many controllers using the same view model.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/confirmation/ConfirmationPaneCommonController.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    @FXML
+    private void closeProperties(ActionEvent event) {
+        viewModel.sendNotification();
+    }
+
+    @FXML
+    public void initialize() {
+        confirmationTitle.textProperty().bind(viewModel.titleProperty());
+        confirmationMessage.textProperty().bind(viewModel.messageProperty());
```
carldea, line 45: Can we bind the properties defined from the addProperty() calls? (See comment above). 

**Note:** These static fields or property names can be `String`s, `Enum`s or `PropertyIdentifier` objects. 
**A preferred way**: The controller could own an inner public enum such as:
```java
 // properties or field names of a UI form (aka mappings of properties to view controls).
 public enum ConfirmationField { 
     CONFIRM_TITLE, 
     CONFIRM_MESSAGE 
};
 ```

The code above should be:
```java
viewModel.addProperty(CONFIRM_TITLE, "");
viewModel.addProperty(CONFIRM_MESSAGE, "");

confirmationTitle.textProperty().bind(viewModel.getProperty(CONFIRM_TITLE));
confirmationMessage.textProperty().bind(viewModel.getProperty(CONFIRM_MESSAGE));
```

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/confirmation/ConfirmationPaneCommonController.java
carldea: One minor caveat that I didn't realize is the "**Close Property Button**" press. Because of this added complexity this may warrant a custom view model. e.g., code to emit an event to close the properties bump out code.

When using just a `SimpleViewModel` (not custom view model) you could implement things inside the `initialize()` method. However, a cleaner approach would be create a subclass that extends `SimpleViewModel`. In spirit of MVVM with respect to JavaFX **controllers** you want to treat them as views (stupid) and presentation logic should be moved out of the controller and into the view model (within reason). The high-level state of the UI form should be managed by the view model. 

In this custom view model class you can simply make calls inside the constructor to `addProperties()` property names (inner enums can live here). e.g., In the constructor you can do the following:
```java
   addProperty(...);
   addProperty(CLOSE_PROPERTIES, false);
   getBooleanProperty(CLOSE_PROPERTIES).subscribe( isClosed -> {
      if (isClosed) {
         // emit event
         // reset
         reset(); // takes original values and resets all properties.
      } 
   });
``` 
Later on in code (inside of the controller class) when user presses the button:
```java
    @FXML
    private void closeProperties(ActionEvent event) {
        viewModel.setPropertyValue(CLOSE_PROPERTIES, true);
    }
```

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
```diff
@@ -105,25 +107,23 @@ private void setupShowingPanelHandlers() {
 
         editFieldsJfxNode = FXMLMvvmLoader.make(config);
 
-        Config closePropertiesConfig = new Config(this.getClass().getResource("close-properties.fxml"))
-                .addNamedViewModel(new NamedVm("genEditingViewModel", genEditingViewModel));
-        JFXNode<Pane, ClosePropertiesController> closePropsJfxNode = FXMLMvvmLoader.make(closePropertiesConfig);
+        ConfirmationPaneCommonViewModel confirmationViewModel = new ConfirmationPaneCommonViewModel();
+
+        Config closePropertiesConfig = new Config(this.getClass().getResource("/dev/ikm/komet/kview/mvvm/view/confirmation/confirmation-pane-common.fxml"))
```
carldea, line 112: Can the common fxml file be in the same namespace as `ConfirmationPaneCommonController`?
**Reason:** That way the it could look like:
```java
Config closePropertiesConfig = new Config(ConfirmationPaneCommonController.class.getResource("confirmation-pane-common.fxml"));
```

Naming is hard, but I'd prefer the controller to be named: `ConfirmationPaneController`.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
```diff
@@ -105,25 +107,23 @@ private void setupShowingPanelHandlers() {
 
         editFieldsJfxNode = FXMLMvvmLoader.make(config);
 
-        Config closePropertiesConfig = new Config(this.getClass().getResource("close-properties.fxml"))
-                .addNamedViewModel(new NamedVm("genEditingViewModel", genEditingViewModel));
-        JFXNode<Pane, ClosePropertiesController> closePropsJfxNode = FXMLMvvmLoader.make(closePropertiesConfig);
+        ConfirmationPaneCommonViewModel confirmationViewModel = new ConfirmationPaneCommonViewModel();
```
carldea, line 110: Can you remove?
Since you want the framework to supply a view model you don't need to instantiate it here.
Once removed you may want an instance variable to reference the viewModel inside the controller.

```java
confirmationViewModel = jfxNode.getViewModel("viewModel").get();
```

Could you name it `ConfirmationViewModel` assuming you create a custom view model? 
Because this scenario is simply a one-to-one (one view and one view model) it is okay to call it ConfirmationPaneViewModel. But it's always good to make names more abstract in case there are complex scenarios such as (many views- to- one view model)

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
```diff
@@ -105,25 +107,23 @@ private void setupShowingPanelHandlers() {
...
             // "Semantic Details Changed" is displayed when form values are Submitted when in EDIT mode
 
-            // TODO Changed is future, waiting for Figma wireframe
-
-//            if (genEditingViewModel.getPropertyValue(MODE).equals(CREATE)) {
-                closePropertiesController.showSemanticDetailsAdded();
-//            } else {
-//                closePropertiesController.showSemanticDetailsChanged();
-//            }
+            confirmationViewModel.setConfirmationMessage(ConfirmationMessages.SEMANTIC_DETAILS_ADDED);
```
carldea, line 126: Can you call `confirmationViewModel.setPropertyValue(CONFIRMATION_MESSAGE, "the message goes here");`
No need to create a class with all the messages ( `ConfirmationMessages.SEMANTIC_DETAILS_ADDED`).

**Reason:** Since the confirmation controller is reusable across all these kinds of windows (aka Properties bump outs toggle button) message text could be whatever the developer decides. e.g. Lidr window, or Concept window they also have confirmation panes based on their transactions.  Since the genediting property controller owns the event subscriber (listener to adding or updating a semantic) it should own the message to be passed into the (reusable) confirmation pane controller's view model.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/confirmation/ConfirmationPaneCommonViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class ConfirmationPaneCommonViewModel extends SimpleViewModel {
+
+    // properties
+    public static final String NOTIFICATION_TOPIC = "notificationTopic";
+    public static final String NOTIFICATION_EVENT = "notificationEvent";
+
+    private StringProperty title = new SimpleStringProperty();
+    private StringProperty message = new SimpleStringProperty();
+    private ConfirmationMessages confirmationMessage;
```
carldea, line 24: Can we create properties for title and message in the constructor?

Tip: Create a public enum or a separate file of the enum properties names?

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/confirmation/ConfirmationPaneCommonViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+     * Send the notification to the provided topic using the provided event.
+     */
+    public void sendNotification() {
+        var topic = getNotifiicationTopic();
+        var event = getNotificationEvent();
+
+        if (topic != null && event != null) {
+            EvtBusFactory.getDefaultEvtBus().publish(topic, event);
+        }
+    }
```
carldea, line 65: See the first comment above ("**Close Property Button**") relating to a new property called `CLOSE_PROPERTIES` set to `true` or `false`.
At this point you'll want to know who is the subscriber of the topic. Since the PropertiesController owns the close properties listener you may want to add the handler code to the view model's `CLOSE_PROPERTIES` property.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/confirmation/ConfirmationText.java
```diff
@@ -0,0 +1,13 @@
+package dev.ikm.komet.kview.mvvm.viewmodel.confirmation;
+
+/**
+ * Contains the confirmation title and message text for display on the confirmation pane.
+ * @param title
+ * @param message
+ */
+record ConfirmationText(String title, String message) {
```
carldea, line 8: Can we remove this record class?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PropertiesController.java
```diff
@@ -305,17 +308,17 @@ private void initialize() {
...
+                confirmationViewModel.setConfirmationMessage(ConfirmationMessages.DEFINITION_ADDED);
             }
             patternViewModel.save();
         };
         EvtBusFactory.getDefaultEvtBus().subscribe(getPatternTopic(), PropertyPanelEvent.class, showPropertyPanelSubscriber);
 
         patternDescriptionEventSubscriber = evt -> {
             if (evt.getEventType() == PatternDescriptionEvent.PATTERN_ADD_FQN) {
-                confirmationController.showFqnAdded();
+                confirmationViewModel.setConfirmationMessage(ConfirmationMessages.FULLY_QUALIFIED_NAME_ADDED);
```
carldea, line 319: Can we use the properties defined?
```java
confirmationViewModel.setPropertyValue(CONFIRM_MESSAGE, "Fully Qualified Name Added");
```

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PropertiesController.java
```diff
@@ -305,17 +308,17 @@ private void initialize() {
...
             } else if (evt.getEventType() == PatternDescriptionEvent.PATTERN_ADD_OTHER_NAME) {
-                confirmationController.showOtherNameAdded();
+                confirmationViewModel.setConfirmationMessage(ConfirmationMessages.OTHER_NAME_ADDED);
```
carldea, line 321: Can we use the properties defined?
```java
confirmationViewModel.setPropertyValue(CONFIRM_MESSAGE, "Other Name Added");
```

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/confirmation/ConfirmationPaneCommonViewModel.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final String NOTIFICATION_TOPIC = "notificationTopic";
+    public static final String NOTIFICATION_EVENT = "notificationEvent";
+
+    private StringProperty title = new SimpleStringProperty();
+    private StringProperty message = new SimpleStringProperty();
+    private ConfirmationMessages confirmationMessage;
+
+    public ConfirmationPaneCommonViewModel() {
+        addProperty(NOTIFICATION_TOPIC, (UUID) null);
+        addProperty(NOTIFICATION_EVENT, (Evt) null);
```
carldea, line 28: Can we create properties for Title, Message, and Close Properties?

kview/src/main/java/module-info.java
```diff
@@ -143,6 +143,11 @@
     opens dev.ikm.komet.kview.klwindows to javafx.fxml, org.carlfx.cognitive;
     opens dev.ikm.komet.kview.mvvm.viewmodel to dev.ikm.komet.application, javafx.fxml, org.carlfx.cognitive;
 
+    exports dev.ikm.komet.kview.mvvm.view.confirmation;
+    opens dev.ikm.komet.kview.mvvm.view.confirmation;
```
carldea, line 147: Can we remove line 147?
Opens keyword is allowing any class to use reflection. This should not be open to the everyone.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/PropertiesController.java
```diff
@@ -105,25 +107,21 @@ private void setupShowingPanelHandlers() {
...
 
-        Config closePropertiesConfig = new Config(this.getClass().getResource("close-properties.fxml"))
-                .addNamedViewModel(new NamedVm("genEditingViewModel", genEditingViewModel));
-        JFXNode<Pane, ClosePropertiesController> closePropsJfxNode = FXMLMvvmLoader.make(closePropertiesConfig);
+        Config closePropertiesConfig = new Config(this.getClass().getResource("/dev/ikm/komet/kview/mvvm/view/confirmation/confirmation-pane-common.fxml"));
+        JFXNode<Pane, ConfirmationPaneCommonController> closePropsJfxNode = FXMLMvvmLoader.make(closePropertiesConfig);
         closePropsPane = closePropsJfxNode.node();
-        closePropertiesController = closePropsJfxNode.controller();
+
+        ConfirmationPaneCommonViewModel confirmationViewModel = (ConfirmationPaneCommonViewModel) closePropsJfxNode.getViewModel("viewModel").get();
```
carldea, line 114: Can you make the call more abstract? No need to downcast.
```java
ViewModel confirmationViewModel = closePropsJfxNode.getViewModel("viewModel").get()
```

jdsmithsos:
I want to make the case of using convenience methods in the view model class to provide the access to the properties.  I added detailed Javadoc to the view model and controller classes, and added the different usage examples in the code with the convenience methods and directly to the properties.

PR #510, opened 2025-06-03
Feature/Finished/IIA-2116: Reset form does not revert back to last committed value when user removes Image by clicking X button

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2116


carldea (approved):
Looks good to me!

PR #512, opened 2025-06-04
Feature/finished/iia 2115 fix for exception getting triggered when user removes concept from c list and c set

swaroopsalvi:
JIRA Fix for https://ikmdev.atlassian.net/browse/IIA-2115

Before deleting: 
image

After Deleting: 
image

Rearrange the Set and List:
 
image


dholubek (approved):
looks good, good work

PR #516, opened 2025-06-05
Feature/Finished/IIA-2139: Dragged Concept from New search field gets populated beside Pattern/Semantic window instead of getting dropped in empty search field

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2139


carldea (approved):
Glad to see this logic back in. I did notice this removed.

Thank you for resolving this issue.

PR #517, opened 2025-06-06
Feature/Finished/IIA-2153: [Autosave] Exception is getting triggered when user removes last entry in CSET/CLIST and reopen the Semantic Window

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2153


carldea (approved):
Looks good to me.

I know this is an urgent fix. I'm still not sure how this filters a nid of zero. We'll have to go over the details during our dev meeting.

PR #529, opened 2025-06-11
Feature/finished/iia 2083 refactor closeproperties confirmation with mvvm 2

jdsmithsos:
Using static imports for new ConfirmationPaneViewModel.
This is the only change for this PR.


carldea (approved):
looks good to me!

PR #538, opened 2025-06-18
Feature/finished/ikdt 1016 change return value of save state

dettmoney:


dholubek (approved):
did some smoke tests.  looks good

PR #544, opened 2025-06-20
Feature/finished/iia 2258 exception triggered on editing existing pattern field causes purpose and meaning fields to disappear

swaroopsalvi:
image


dholubek (approved):
looks good

PR #546, opened 2025-06-20
Feature/finished/iia 2266 import image fails

dukke:
Fixes https://ikmdev.atlassian.net/browse/IIA-2266


swaroopsalvi (approved):
Looks good.

PR #550, opened 2025-06-23
Feature/finished/iia 2257 tab navigation keypress once

bhharsh13:
Fixed the Number of Tab Jumps Needed to Navigate Editable Fields in Semantic Window.

https://ikmdev.atlassian.net/browse/IIA-2257

Summary of Changes :

Set setFocusTraversable as false at the control levels for KLStringControl, KLIntegerControl, KLFloatControl, KLBooleanControl

Validated the changes on local.

image


bhharsh13:
This has dependency on https://ikmdev.atlassian.net/browse/IIA-2267 for the cset and clist tab navigation. Once the focus is added then we can fix tab navigation. Hence holding this PR till then .

bhharsh13:
Story AC is modified to include String/Integer/Float/Boolean editable fields in a Semantic Window. C set and C List is moved to https://ikmdev.atlassian.net/browse/IIA-2267

jdsmithsos (commented):
For the KLFloat, Integer, StringControlSkin classes, they are the only classes that extend the KLDebounceControlSkin.  For these specific classes, the call to control.setFocusTraversable(false) could be moved up to the common base class.

The KLBooleanControlSkin doesn't extend from KLDebounceControlSkin, so the call must remain there.

dukke (approved):
Looks good to me!

@jdsmithsos made a valid comment about setting the focus tranversable in the "DEBounce..." class for Integer, Float and String.

I'd also add to that, but maybe this would be for another ticket, that we could create a base class for all editable controls (there is some common behavior among them). When we do that, we could simply do this change in just the "KLEditableBaseControl" class. @carldea would love to know your thoughts on this...

We could also create a base class for all controls (editable and readonly) because there's also common things among them (all controls have a title for example). 


Approved.

bhharsh13:
Made the changes to include FocusTraversable property in the KLDebounceControlSkin for String, Integer and Float. Validated the changes and looks good. @jdsmithsos , please help to approve.

jdsmithsos (approved):
looks great!

PR #556, opened 2025-06-26
Feature/observables

dholubek:
updates to KlView class


carldea (approved):
This looks good to me. I think there is more to come with respect to klwindow services and interfaces. This would be the start of utilizing the KlContext and Node properties capability. We would need to refactor or repurpose @jdsmithsos view coordinate effort.

PR #557, opened 2025-06-26
Feature/finished/jpro 136 persist properties panel state across journal window sessions

indritbeqiri:
**JIRA Ticket**: [JPRO-136](https://ikmdev.atlassian.net/browse/JPRO-136)

**Details**: Related before and after videos are attached to the Jira ticket.

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/klwindows/EntityKlWindowState.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/klwindows/EntityKlWindowState.java
@@ -16,13 +16,16 @@
 package dev.ikm.komet.kview.klwindows;
 
 import dev.ikm.komet.preferences.KometPreferences;
+import org.eclipse.collections.api.map.MutableMap;
+import org.eclipse.collections.impl.factory.Maps;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.Arrays;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.prefs.BackingStoreException;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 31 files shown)


kview/src/main/java/dev/ikm/komet/kview/klwindows/EntityKlWindowState.java
```diff
@@ -762,28 +761,123 @@ private static void loadEntityProperties(KometPreferences preferences, EntityKlW
      * It attempts to load each property with its appropriate type.
      *
      * @param preferences the preferences node to load from
-     * @param state the window state instance to populate
+     * @param state       the window state instance to populate
      */
-    private static void loadAdditionalProperties(KometPreferences preferences, EntityKlWindowState state) {
+    private static void loadAdditionalProperties(KometPreferences preferences, EntityKlWindowState state)
```
carldea, line 766: @indritbeqiri Can you put a comment and TODO statements to mention in the near future to leverage Knoweledge Layout module's `KlContext` facility? Also, I didn't get a chance to share Komet's framework has additional preference value type persistence APIs we should be leveraging. I think some of these code/functions maybe redundant. Would like a spike or analysis on persistence of values and conventions.

@kec we would like to get JPro's PR merged as a working version (1st cut/stab) of persisting other value types and preserving a chapter window when the user has opened their respective Property toggle bump outs (taking the available width). We will need to create another ticket to refactor code to use our framework apis/interfaces as it relates to **knowledge layout**.

indritbeqiri: @carldea I just added a TODO statement to the `EntityKlWindowState` class since we don't yet know the full scope of the mentioned refactoring to leverage the `KlContext` facility.

PR #559, opened 2025-06-27
Feature/finished/iia 2304 author concept exception

dholubek:
default to public id string when the highlighted string comes back as null

image


framework/src/main/java/dev/ikm/komet/framework/search/SearchResultCell.java
```diff
@@ -73,7 +73,10 @@ public void updateItem(Object item, boolean empty) {
             if (item instanceof LatestVersionSearchResult latestVersionSearchResult) {
                 TextFlow textFlow = newTextFlow();
 
-                String matchedText = latestVersionSearchResult.highlightedString();
+                String matchedText = "";
+                if (latestVersionSearchResult.highlightedString() != null) {
+                    matchedText = latestVersionSearchResult.highlightedString();
+                }
```
carldea, line 79: Can we use the public id string instead of an empty string?

Reasoning: When displayed to the user the semantic (row) is blank. We should show something.

dholubek: fixed

PR #563, opened 2025-07-03
Feature/finished/iia 2374 improve search popup UI

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2374

## Before the change
image

## After the change (in this PR):
image


carldea (approved):
Excellent work @dukke !!!
I encountered this issue (radio button dot is not centered). Glad you figured it out! 
```java
    -fx-background-radius: 1.0em; /* large value to make sure this remains circular */
    -fx-padding: 0.333333em; /* 4 -- radius of the inner black dot when selected */
```

PR #564, opened 2025-07-04
Feature/finished/iia 2192 lifehash identicons

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2192

--------------

## Changes in this PR

This PR replaces the old algorithm for generating identicons with a new **Lifehash algorithm**.

### How it looked before
image

### How it looks with this PR
The Image below is using Lifehash **Detailed mode**.
image

### Considerations

The library that was in the POC is no longer existing. I'm using a library that is in Maven Central that is the Java implementation of the Lifehash algorithm. It's called "toucan".

Lifehash supports different modes for generating the Lifehash identicons. Here's my take on each mode:
- Version2: This is an OK looking algorithm, it's also the fastest. Pressing to show the Pattern Browser takes **2 seconds** in this mode.
- Detailed: IMO this produces the best looking icons. It takes longer to generate the icons than Version2. Pressing to show the Pattern Browser takes **4 seconds** in this mode.
- Fiducial: This looks better than Version2 but the icons look like water color paintings. IMO Detailed mode produces icons that are better suited for Komet. This is also the mode that seems to take longer to render. Pressing to show the Pattern Browser takes **5 seconds** in this mode.
- Version1: Seems to be as fast as Version2.
- Grayscale: We likely want color so I'm not considering this mode.

In this PR ~~I'm using **"Detailed"** mode~~ I'm using **Version2**. I think we can change it later to **Detailed mode** once we improve identicon generation performance.
I'm using the algorithm for copying the bytes into a JavaFX Image that was in the POC. That way it saves an extra BufferedImage from being created.

### Possible next steps after merging

We can improve icon generation:
1. Render the icons in a background thread and have a callback that gets called when it's finished
2. Cache the icon results. If the icon has already been generated returned the cached icon
3. There might be something in the code of the Pattern Browser that we can improve (performance wise).
4. IMO the icons might look better if we round the corners just a bit. In that case we would probably return a Node rather than an Image.

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.53.0-IIA-2192-Lifehash-Identicons-SNAPSHOT</version>
+        <version>1.53.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.53.0-SNAPSHOT</version>
+        <version>1.53.0-IIA-2192-Lifehash-Identicons-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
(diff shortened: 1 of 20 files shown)


carldea (approved):
Great work!
I believe this implementation is good for now as an improvement from the previous algorithm to display unique images based on a components `PublicId`. 

In the future we would like this to be service oriented to allow pluggable implementations of different identicon algorithms to allow the user to choose from.

PR #565, opened 2025-07-07
Feature/finished/iia 2245 new naming pattern for semantics

dholubek:
change description name, it is reused already in the tooltip.

image

after updating with tooltip changes from main

image


bhharsh13 (approved):
Looks good to me

dholubek:
moving to draft... after updating main to the feature branch the tooltip stuff is breaking in my feature branch... will investigate

dukke (approved):
Looks good to me. Nice work, I like that the big if got removed :)

PR #572, opened 2025-07-09
Feature/finished/iia 2306 concept not populated via context menu in new search and semantic details

swaroopsalvi:
Fixed https://ikmdev.atlassian.net/browse/IIA-2306


carldea (approved):
@swaroopsalvi Thanks for our dev discussion on how this could be improved. I think your approach is good for now. Since the callback (consumer) into the menu option to summon a concept window is layers deep. We can revisit the work and refactor.

PR #573, opened 2025-07-10
select concept when doing a dnd operation, instead of highlighting it

jperedadnr:


carldea (approved):
@jperedadnr  Looks good to me.

PR #574, opened 2025-07-10
feature/finished/IIA-2384: [TE Feedback] Add Semantic Field Values to Semantic Tooltip in Pattern Browser for Image data types

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2384

------------------

Adds tooltip for Image data types
image


kview/src/main/java/dev/ikm/komet/kview/klfields/KlFieldUtils.java
```diff
@@ -0,0 +1,33 @@
+package dev.ikm.komet.kview.klfields;
```
carldea, line 1: Can we move these static utility functions to `KlFieldHelper.java`?

PR #576, opened 2025-07-10
Passed journalTopic to limit event in Journal Window.

swaroopsalvi:
https://ikmdev.atlassian.net/browse/IIA-2405


carldea (approved):
@swaroopsalvi  Looks good to me!

PR #583, opened 2025-07-15
Feature/finished/IIA-2384: semantic tooltip image field AND IIA-2383: Semantic Tooltip for Concept, CSet and CList data types

dukke:
**The other PR got messed up. Creating this new one with the change that Carl requested to reduce the size of the image in the semantic tooltip preview.**

Fixes: 
1. https://ikmdev.atlassian.net/browse/IIA-2383
2. https://ikmdev.atlassian.net/browse/IIA-2384
----------

Adds a tooltip to semantics with Image Field type and Component, CSet and CList field type

image

The image field below is actually in reality half the size.
image


bhharsh13 (approved):
Looks good to me.

PR #588, opened 2025-07-16
Add subscriber to concept navigator to detect changes in dataset

jperedadnr:
after import or axioms.
This PR covers two different scenarios, and each has its own solution, but in the end, the ConceptNavigator treeView gets refreshed, and the selection (if any) is still visible.

The first scenario (the export/import dataset case) uses a subscriber to `RefreshCalculatorCacheEvent`, while the second scenario (concept axioms updated) uses the weak subscriber to `Entity::provider`.


framework/src/main/java/dev/ikm/komet/framework/observable/ObservableEntity.java
```diff
@@ -224,7 +224,11 @@ private static class EntityChangeSubscriber implements Subscriber<Integer> {
         public void onNext(Integer nid) {
             // Do nothing with item, but request another...
             if (SINGLETONS.containsKey(nid)) {
-                get(nid);
+                if (Platform.isFxApplicationThread()) {
+                    get(nid);
+                } else {
+                    Platform.runLater(() -> get(nid));
+                }
```
carldea, line 231: I believe you have a merge conflict with this file. 
I believe the main branch does the same thing as above if that make any sense.

PR #589, opened 2025-07-17
Bugfix/iia 0000 pattern versions defect part1

carldea:
**Scenarios to fix:**

1. Creation of too many pattern versions – Done
2. Creation of multiple FQN semantics – Done
3. Creation of multiple Other Name semantics – Done
4. Adding US Dialect semantic to FQN – Done
5. Step 3 plus Edit FQN - Done
6. Step 2 plus Add other name - Done.
7. Add other name 2, click on label to edit then hit done, click on publish. should get one version. - Done 
8. After a successful publish identicon should be updated  - Done
9. Identifier is now able to be copied - Done

**NOTE:** **It is safe to press the globe (publish) button again & again** :-). If you press publish button subsequently with an edit FQN and add/edit Other Names the database will update the semantics correctly. **It is safe to press again** if nothing has changed (no changes will be written to the database). However, the definition cannot be changed once you've initially pressed publish. This will be fixed later.

**Outstanding issue:**
- When editing a pattern's definition (meaning & purpose) a subsequent publish press does not create a new pattern version.
- After a newly created Pattern window, closing journal view should save the preference. After restart. Workaround is just resummon window.


rbradley813 (approved):

swaroopsalvi (commented):
looks good.

PR #593, opened 2025-07-21
Show search results in concept navigator control

jperedadnr:
Search results from the Next Gen Search tab can be shown in the Concept Navigator, via context menu -> Open in Concept Navigator.

This PR updates the subscriber to use the NextGen ConceptNavigator control instead of the classic one.

Given that there is no information from the parent of the concept from the search result, the shorter path is selected from all possible parents of this concept.


carldea (approved):
This looks good to me!

PR #594, opened 2025-07-21
IIA-2432 FQN Date format display in Pattern Window

bhharsh13:
Fixed the display for FQN Date format in Pattern Window - https://ikmdev.atlassian.net/browse/IIA-2432

Please help to review the PR.
Note: Primordial would be displayed if the date matches Long.MIN_VALUE + 1.

Before the Change :
image (2)

After the Change:

Primordial Date for Long MIN_VALUE+1

This is for the New Pattern created.
Newly created Pattern


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -469,12 +469,19 @@ public void onChanged(Change<? extends DescrName> change) {
...
+                                .get().stamp().time();
+
+                        //Check if the timestamp is valid. If it matches Long.MIN_VALUE + 1
+                        if (rawTime == Long.MIN_VALUE + 1) {
+                            return "Premundane";
+                        }
+
+                        Instant stampInstance = Instant.ofEpochSecond(rawTime);
+                        ZonedDateTime stampTime = ZonedDateTime.ofInstant(stampInstance, ZoneOffset.UTC);
+                        String fqnDateAddedStr = DateTimeFormatter.ofPattern("MMM dd, yyyy").format(stampTime);
```
dukke, line 483: Do we want to hard code the month first and then the day and then the year?
Some countries put the day first, and then the month and then the year... perhaps we should grab the current locale being used (could be spanish local, portuguse local, etc) and format it using that locale?

cc @carldea

dukke (changes requested):

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
```diff
@@ -469,12 +469,19 @@ public void onChanged(Change<? extends DescrName> change) {
...
+                                .getFullyQualifiedDescription(patternProperty.get().nid())
+                                .get().stamp().time();
+
+                        //Check if the timestamp is valid. If it matches Long.MIN_VALUE + 1
+                        if (rawTime == Long.MIN_VALUE + 1) {
+                            return "Premundane";
+                        }
+
+                        LocalDate localDate = Instant.ofEpochSecond(rawTime).atZone(ZoneId.systemDefault()).toLocalDate();
+                        String fqnDateAddedStr = localDate.format(DateTimeFormatter.ofPattern("MMM d, yyyy"));
```
dukke, line 482: So the problem is the hardcoded date format:
"MMM d, yyyy"
This way we'll always get the month first and then the day and then the year. We want something that's not hardcode like that.  Different countries can have it differently, like there are some where dates are shown as the day first, then the month, then the year.

I think we need some way of grabbing the current local and using that.. though I'm not entirely sure how much of a priority is that 
cc @carldea

dukke (changes requested):
Left a comment, I think we're still hard coding the way we present the date.

Thanks!

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
carldea: @bhharsh13 I agree with Pedro @dukke .

Can we try the following:
```java
// Java 8 DateTimeFormatter
DateTimeFormatter.ofLocalizedDate(FormatStyle.SHORT).withLocale(locale);
DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM).withLocale(locale);
```

bhharsh13:
Hi @carldea ,
Please help to approve the PR for the FQNDateAdded displaying Premundane.
fqnDate Dsiplay fix

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternDetailsController.java
carldea: @dukke 
I worked with @bhharsh13  to format the date based on your local and now will format correctly.

dukke (changes requested):
@bhharsh13 @carldea 
If the last screenshot from @bhharsh13 is how it looks after this PR, there's a regression.

We should fix the regressions first, this happened because of errors when merging. I had this too and I had to fix it afterwards when I merged my PR which was in the same code base.

@bhharsh13 go through the tickets that got merged recently, which got this to have the merge conflict (the tickets that changed code in these same classes) and ensure everything on those tickets is still working, if you can please.

bhharsh13:
Hey Pedro Duque-Vieira, I have pulled the latest code merged and tried to test it with my changes before merging the PR. The changes looks good and not conflicting since I don't see anyone making changes in PatternDetailsController/ PatternViewModel. CSS styling also looks good.  We should be good to merge the PR.
image

PR #602, opened 2025-07-23
Feature/finished/iia 2501 bugfix retired label

dholubek:
image

previously the retired label was getting removed when you clicked the search filter options

image


bhharsh13 (approved):
Looks good to me

PR #606, opened 2025-07-24
Feature/finished/ar 000 properly create membership semantics

rbradley813:


carldea (approved):
Looks good to me. 
I'm assuming we've tested patterns with field definitions just in case.

kec:
If this resolves the issue for now, super. From the semantic record, and the transaction perspective, it should not matter if there are no fields or not. The process for committing them and writing them is the same. There is still a commit process for the stamp, which includes the stamp fields.

kec (approved):

PR #608, opened 2025-07-28
Feature/finished/ar 000 expand hierarchy fx thread fix

rbradley813:


carldea (approved):
Looks good to me!
This is good for now.
We can check later in other places.

PR #601, opened 2025-07-23
IIA-2437, IIA-2438 Remove the Apply filter option from NextGen Search and NextGen Navigator Filter Menu

bhharsh13:
Remove "Apply" Button From NextGen Search and NextGen Navigator Filter Menu - https://ikmdev.atlassian.net/browse/IIA-2437
https://ikmdev.atlassian.net/browse/IIA-2438

**Before Fix:**
Screenshot 2025-07-15 at 10 32 46 PM

**After Fix:** 

**NextGenSearch Filter-**

Remove Apply Button

**NextGen Navigator Filter -**

Remove Apply Button NextGenNavigator

PR #613, opened 2025-07-30
Update filter options property whenever an option is changed

jperedadnr:
by the user interaction.

This allows detecting any changes in the filter options, without waiting for the user to press the Apply button.

The Apply button remains enabled, but it won't do anything other than collapsing the options, since the changes are already set.


carldea (approved):
Looks good to me!

Later on we can remove the apply button. 
@bhharsh13 's PR here: https://github.com/ikmdev/komet/pull/601

PR #616, opened 2025-08-01
IIA-1586 Border of the Expanded Pattern should get highlighted

bhharsh13:
https://ikmdev.atlassian.net/browse/IIA-1586 - When user click on the Pattern drop down which has semantics, the border of the expanded pattern should get highlighted. Increased the border width and the bgColor

**Before Change:**

**After Change:**
Pattern Highlighting


jdsmithsos (approved):
looks good

PR #620, opened 2025-08-04
Feature/finished/iia 2535 bugfix pattern default concept

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2535

tested with the attached dataset and the next gen pattern navigator loads fine.


carldea (approved):
Looks great to me!

I trust it is @rbradley813 approved. ;-)

PR #628, opened 2025-08-06
Feature/finished/iia 2499 pattern activate inactivate classic

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2499

image


carldea (approved):
Excellent work!

PR #632, opened 2025-08-07
Improve ConceptNavigator performance

jperedadnr:
This PR fixes some performance issues with datasets like GUDID, by introducing three changes:
- cache `Navigator::isLeaf` calls
- fetch children for a given treeItem in a background thread
- prune collapsed branches to reduce model in memory

There are some side effects now, as the process is not synchronous anymore, and visual hints will be needed, as a follow-up, to show that some tasks are running without immediate UI results (for instance, expanding the treeView for a search result with a huge number of children).


carldea (approved):
@jperedadnr  Excellent work! Looks good to me. I will approve and merge.
This control is getting better and better.

PR #633, opened 2025-08-07
IIA-2381  Remove Extra Space Between Lines in Concept, Pattern and Semantic Window Header

bhharsh13:
Removed Extra Space Between Lines in Concept, Pattern and Semantic Window Header - https://ikmdev.atlassian.net/browse/IIA-2381
Please help to approve the PR.

**Before Change:**

**After Change:**
Concept Header
Pattern Header
Semantic Header

### DIFF
--- a/kview/src/main/resources/dev/ikm/komet/kview/controls/public-id-control.css
+++ b/kview/src/main/resources/dev/ikm/komet/kview/controls/public-id-control.css
@@ -20,7 +20,7 @@
 }
 
 .public-id > .title-label {
-    -fx-font-family: "Noto Sans Bold";
+    -fx-font-family: "System Bold";
     -fx-font-size: inherit;
 }
 
--- a/kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
+++ b/kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
@@ -153,6 +153,7 @@
     -fx-font-family: "Noto Sans";
     -fx-font-size: 11px;
     -fx-fill: -Grey-11;
+    -fx-font-weight: bold;
 }
 
 .stamp-label {
@@ -218,6 +219,11 @@
     -fx-alignment: top-left;
     -fx-max-width: infinity;
     -fx-max-height: infinity;
+}
+
+.concept-definition-title{
+    -fx-text-fill: -Grey-11;
+    -fx-font-weight: bold;
 }
 
 .titled-pane {
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.54.0-SNAPSHOT</version>
+        <version>1.54.0-IIA-2381-remove-extra-space-concepts-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.54.0-IIA-2381-remove-extra-space-concepts-SNAPSHOT</version>
+        <version>1.54.0-SNAPSHOT</version>
     </parent>
     <artifactId>application</artifactId>
     <dependencies>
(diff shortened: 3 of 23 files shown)


bhharsh13:
@carldea ,
Title Pane is aligned with the Stamp module. Below that are the Other details displayed.

dukke (changes requested):
I noticed a few things:
1 - Concept Window: Identifer public id text changed to bold... do we want that?
2 - Concept Window: What exactly was fixed?
3 - Definition text is being cut out.. (maybe could be a separate ticket to fix this)
4-  Semantic Window: There's still quite a good amount of space between the title and the text below it.

I think the fix for this would probably be redoing the whole way we lay out the top left part of the window (title, icon, text below title, etc).....

bhharsh13:
In the Concept window, there was extra space between Definition and IDENTIFIER. 
@dukke , I agree we need to completely redo the layouts.

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/details/DetailsController.java
```diff
@@ -900,6 +900,7 @@ private void setupDisplayUUID(EntityFacade entityFacade, ViewCalculator viewCalc
         String idStr = String.join(", ", idList);
 
         identifierControl.setPublicId(idStr);
+        identifierControl.setStyle("-fx-font-weight: normal;");
```
dukke, line 903: can we set this in the css file rather than like this? Supposedly the identifier control text (not the label that goes left) is already NOT bold.

kview/src/main/resources/dev/ikm/komet/kview/controls/public-id-control.css
```diff
@@ -3,10 +3,9 @@
        where the height is not changed as the mouse cursor enters and exits
        the control.  30px seems to work well for the current settings for
        fontSize and the icon scale. */
-    -public-id-height: 30px;
 
-    -fx-pref-height: -public-id-height;
-    -fx-min-height: -public-id-height;
+    -fx-pref-height: 22px;
+    -fx-min-height: 22px;
```
dukke, line 8: can we use the variable like we were using. That is: 

    -public-id-height: 22px;

    -fx-pref-height: -public-id-height;
    -fx-min-height: -public-id-height;

kview/src/main/resources/dev/ikm/komet/kview/controls/public-id-control.css
```diff
@@ -20,8 +19,11 @@
 }
 
 .public-id > .title-label {
-    -fx-font-family: "Noto Sans Bold";
-    -fx-font-size: inherit;
+   -fx-font-family: "Noto Sans Medium";
+   -fx-font-weight: 500;
+   -fx-font-size: 12px;
+   -fx-text-fill: black;
+   -fx-font-weight: bold;
```
dukke, line 26: I think that unfortunately this properties in javafx css might not actually work. 
Set the font title instead of using the property. 
Like if you want a bold text use:
-fx-font-family: "Noto Sans Bold";

If you want normal text use:
-fx-font-family: "Noto Sans";

Semi bold (which is a step before bold):
-fx-font-family: "Noto Sans Semibold";

The font files must exist though...

kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
```diff
@@ -153,6 +153,7 @@
     -fx-font-family: "Noto Sans";
     -fx-font-size: 11px;
     -fx-fill: -Grey-11;
+    -fx-font-weight: bold;
```
dukke, line 156: I don't think this works. Like I said earlier.

kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
```diff
@@ -220,6 +221,14 @@
     -fx-max-height: infinity;
 }
 
+.concept-definition-title {
+    -fx-font-family: "Noto Sans Medium";
+    -fx-font-weight: bold;
```
dukke, line 226: I don't think this works. Like I said earlier.

dukke (changes requested):
Great work @bhharsh13 

There's some small change I think would be good to do.

Also the identifier text needs to NOT be bold.

kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
bhharsh13: modified to use the -fx-font-family for bold.

PR #634, opened 2025-08-08
Add initial value to property to prevent NPE

jperedadnr:
After the changes in #619 , when the initialFilterOptionsProperty gets queried, as it doesn't have an initial value, causes a NPE.

This PR is a quick fix to avoid such NPE, but typically the initial value should be set via setter.


dholubek (approved):
approved

PR #635, opened 2025-08-08
IIA-2560: [Concept Creation] Null Pointer Exception When Adding Other Name During Concept Creation

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2560


dholubek (approved):
approved

PR #642, opened 2025-08-11
JPRO-162: Sign In User author, and App/WebApp refactoring

FlorianKirmaier:
In this PR, we achieved two major improvements.

- The Sign In Page for the Author Selection is now working. (JPRO-162)[https://ikmdev.atlassian.net/browse/JPRO-162]
- The App and WebApp are refactored, avoiding code duplication and simplifying integration.

SignInPage:
The new LoginAuthorSelection page allows to select a user.
Currently, it’s using fake data. Later, it will use the real users from the database.

WebApp/App Merge:

We did this refactoring now to avoid integrating the new LoginPage twice. This gives a lot of improvements.

These changes don't alter anything functionally, but they provide numerous improvements.

- Removed about 1.5k lines of duplicated code.
- Makes onboarding for the main class much easier, which also triggered this change.
- The code is now separated into several files, making it easier to find the correct place.


carldea:
> @carldea Thank you for the positive feedback! I’ve merged and updated the pom to 1.55.0-SNAPSHOT. It would be great if we could set it up, that not every commit on my side requires approval. I totally agree - testing is very important here, because a lot of code has been touched.

We are working on the permission issue. We are still in the process of QA testing :-)

carldea (approved):
Looks great!
I will approve for now. Until the QA testing is finished I will merge this effort first (tomorrow).
The other PRs pending will be next to be merged.

PR #646, opened 2025-08-12
changes for KOMET interrupted exception on export process cancel button

casaras1:


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/changeset/ExportController.java
```diff
@@ -293,22 +294,41 @@ private void performChangeSetExport(final FileSavePicker fileSavePicker, final l
...
+                        delete = true;
+                    } else {
+                        LOG.error("Export to file '{}' failed", exportFile, e);
+                        delete = true;
+                    }
+                    throw new CompletionException(e);
+                } finally {
+                    if (delete) {
+                        deleteFile(exportFile);
+                    }
```
carldea, line 326: Nice work on handling every scenario however, this can be handled easily with a `CompletableFuture`.
Can you make the code more concise by using `CompletableFuture.whenComplete()` or `.handle()`? 

You probably didn't know but `ProgressHelper.progress()` returns a `CompletableFuture`.

The code could/should look like the following:
```java
fileSavePicker.setOnFileSelected(exportFile -> {
    if (exportFile == null) {
        LOG.warn("Export file is null");
        closeDialog();
        return;
    }
    closeDialog();

    CompletableFuture<Void> exportFuture = ProgressHelper.progress(exportEntities, "Cancel Export");

    exportFuture.handle((result, throwable) -> {
        if (throwable != null) {
            LOG.error("Export to file '{}' failed", exportFile, throwable);
            deleteFile(exportFile);
        } else {
            LOG.info("Export completed successfully to file {}", exportFile);
        }
        return null;
    });
});
```

casaras1: change is done.

PR #652, opened 2025-08-13
feature/finished/IIA-2429-2212-fixed-obslangcoord-nooverride-withoverride

jdsmithsos:
This PR fixes two of JD's bug tickets:

https://ikmdev.atlassian.net/browse/IIA-2429
https://ikmdev.atlassian.net/browse/IIA-2212

Also fixes two of Swaroop's bug tickets:

https://ikmdev.atlassian.net/browse/IIA-2344
https://ikmdev.atlassian.net/browse/IIA-2439

And probably other bugs that are in the Language area of the view coordinate.

Summary of changes:

- Now every call in the Komet app to ObservableViewNoOverride.makeOverridableViewProperties() has a unique name, so each view property can be easily identified and breakpoints can be set based on the unique name
- Carl contributed to this PR by changes to create a new overridable view property rather than using an existing one, and ensures that the Journal window's view coordinates are used as the parent
- Fixes were made to
    - ObservableViewLanguageCoordinateNoOverride
        - using the incorrect properties for dialectPattern, was not using descriptionPattern correctly
    - ObservableViewLanguageCoordinateWithOverride
        - missing the make method for descriptionPattern
        - using the incorrect properties for dialectPattern and descriptionPattern

Before changes:

- The first time that the Journal parent view coordinates Language change description to FQN happened, an exception occurred
- Selecting the language to Spanish didn't propagate to the child view coordinates

After changes:

- No exception occurs the first time
- The language selection to Spanish does propagate to the child view coordinates
    - But the text is messed up.  Needs to be addressed in a different bug ticket.

image

Spanish propagates, but the text is messed up.  (address in a different ticket)

image


carldea (approved):
Excellent work JD!
Great work in investigating, fixing, and tracking down all the children needing a derived `ViewProperties` (`WithOveride`). Nice to see the propagation code work.  
We should ask about the name() method to be sure no other abstraction or API might be expecting the default name called `"View"`.

PR #653, opened 2025-08-14
feature/finished/IIA-2585-fixed-journal-view-coordinate-menu-from-jpro-changes

jdsmithsos:
Emergency fix for the Journal window's view coordinate menu.
Added the required call to JournalWindow.setup(), which initializes the view coordinate menu.


carldea (approved):
Looks good, good catch.

PR #656, opened 2025-08-18
IIA-2523: Set Default Values for Property Values in Edit Panel for All Subsequent Other Names

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2523

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/OtherNameViewModel.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/OtherNameViewModel.java
@@ -1,0 +1,22 @@
+package dev.ikm.komet.kview.mvvm.viewmodel;
+
+import dev.ikm.tinkar.entity.ConceptEntity;
+
+import static dev.ikm.komet.kview.mvvm.viewmodel.OtherNameViewModel.OtherNameProperties.FQN_CASE_SIGNIFICANCE;
+import static dev.ikm.komet.kview.mvvm.viewmodel.OtherNameViewModel.OtherNameProperties.FQN_LANGUAGE;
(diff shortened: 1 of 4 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/OtherNameViewModel.java
```diff
@@ -0,0 +1,22 @@
+package dev.ikm.komet.kview.mvvm.viewmodel;
+
+import dev.ikm.tinkar.entity.ConceptEntity;
+
+import static dev.ikm.komet.kview.mvvm.viewmodel.OtherNameViewModel.OtherNameProperties.FQN_CASE_SIGNIFICANCE;
+import static dev.ikm.komet.kview.mvvm.viewmodel.OtherNameViewModel.OtherNameProperties.FQN_LANGUAGE;
+import static dev.ikm.komet.kview.mvvm.viewmodel.OtherNameViewModel.OtherNameProperties.HAS_OTHER_NAME;
+
+public class OtherNameViewModel extends DescrNameViewModel {
```
carldea, line 9: Can we just use DescrNameViewModel?  
FQN and Other name are both description semantics. DescrNameViewModel should have a enum of properties. if not we should refactor. If it's out of scope let me know.

dukke: @carldea This new viewmodel extends from DescNameViewModel.
DescrNameViewModel is used for FQN, "other name" and perhaps some other places. 
I thought I'd create a specialized view model (OtherNameViewModel) that extends from DescNameViewModel since the properties I defined here in this view model only make sense for "other name" (the Jira ticket description has the background) - properties like FQN_LANGUAGE. HAS_OTHER_NAME .

Let me know your thoughts...

carldea (approved):
Looks and seems good to me.

PR #657, opened 2025-08-18
Feature/finished/iia 2353 retrieve list of author concepts to select during login to komet

swaroopsalvi:


carldea (approved):
Looks good to me!

PR #658, opened 2025-08-18
IIA-2613 fixed the discovered view coordinate coding errors

jdsmithsos:
During the view coordinate effort, it was discovered that there were coding errors that was causing problems within the Language coordinates.
I fixed those coding errors, which then fixed 5 defects with view coordinate propagation.
I thought it would be a good idea to analyze the rest of the view coordinate classes, to see if there are any additional coding errors.
And there are.

This is a DRAFT PR to communicate what I found, and to get feedback from the team about moving forward with these fixes to merge to the main branch.

There are two other changes that are not related, but need to be done as well, to streamline the prior changes that were made to ObservableViewBase and ObservableViewWithOverride.  Actually, these two changes can be moved to a different PR if needed to not have any riders on the intent of this PR.

Before changes:

Coding errors discovered.

After changes:

Discovered coding errors fixed (hopefully)


carldea:
@jdsmithsos Assuming you've updated, tested your branch you may want to move it out of draft mode.

PR #660, opened 2025-08-19
Refactor FilterOptions and add Any button, add Any and Excluding

jperedadnr:
to the Module option

image

This can be also extended to any other option (and button names can be customised per option as well).


carldea (approved):
Looks good to me!
Nice work. I would like @dholubek  to weigh-in since most of his current effort uses the filteroptions for each view coordinate filters.

dholubek (approved):
I think this looks good.

PR #667, opened 2025-08-21
Feature/finished/iia 2652 change included modules including individual modules updates parent menu

swaroopsalvi:
https://ikmdev.atlassian.net/browse/IIA-2652

image

image

image


carldea (approved):
I approved, On another PR we should refactor and remove the instance variable `windowView`. And places will need to use `windowSettings.getView()`.

PR #670, opened 2025-08-22
Feature/finished/iia 2534 next gen filter inheritance

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2534

initial filter loaded from its parent:
image

about to change from development path to sandbox path...
image

child menu changes:
image

NOTE: the search refresh isn't working yet to reflect the sandbox path or update even if we hit apply; that AC has been separated out of this ticket to get this initial functionality working.


carldea (approved):
Looks good as far as i can tell. 
I think this is a good first stab at syncing view coordinates to be applied to the ui control. I know many of the options are simple Strings. But it would be nice to make use of JavaFX property bindings instead of the copies between parent-view-props to ui popup.

PR #671, opened 2025-08-22
IIA-2652 put back in the journal view coordinate menu listener

jdsmithsos:
Put back in the Journal window view coordinate menu listener.

The real fix for IIA-2652 will come sometime next week...


carldea (approved):
Looks good.

PR #674, opened 2025-08-25
Feature/finished/iia 2623 implement author attribution in concept window concept version form

swaroopsalvi:
image

image

**_Before Change:_**
image

_**After Change:**_
image

**Online User:**

image

**Select DB**
image

**Login using KOMET user**
image

image

image

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/StampViewModel2.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/StampViewModel2.java
@@ -163,11 +163,10 @@ public class StampViewModel2 extends FormViewModel {
         // Choose one item from the Sets as the module and path. Items will use .equals(). STATUS property value is an Enum.
         ConceptEntity module = modules.stream().filter( m -> m.nid() == stampEntity.moduleNid()).findFirst().orElse(null);
         ConceptEntity path = paths.stream().filter( m -> m.nid() == stampEntity.pathNid()).findFirst().orElse(null);
-        ConceptFacade authorConcept = viewProperties.nodeView().editCoordinate().getAuthorForChanges();
 
         setPropertyValue(STATUS, stampEntity.state());
         setPropertyValue(TIME, stampEntity.time());
-        setPropertyValue(AUTHOR, authorConcept);
+        setPropertyValue(AUTHOR, stampEntity.author());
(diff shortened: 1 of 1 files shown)


swaroopsalvi:
> Quote reply

If we show the current logged-in user the in the STAMP form, then then the submitted data will not be the same as the form data, unless we want to update the data when the input values change. Just seems like unnecessary overhead. So values won't be updated in the DB unless the user changes it..

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/properties/StampAddController.java
```diff
@@ -81,7 +81,12 @@ private void initAuthorField() {
 
             @Override
             protected String computeValue() {
-                return "Author";
+                if (getViewProperties() != null) {
+                    ConceptFacade authorConcept = stampViewModel.getPropertyValue(AUTHOR);
+                    return getViewProperties().calculator().getPreferredDescriptionTextWithFallbackOrNid(authorConcept.nid());
```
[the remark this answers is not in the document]
carldea: I think `getPreferredDescriptionTextWithFallbackOrNid()` is the correct (preferred) way. Users are concepts, and concepts have semantics (Description Semantics). And they respond to the parent view properties, and for display purposes it could be in a different language.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/StampViewModel2.java
```diff
@@ -222,6 +222,7 @@ public StampViewModel2 save() {
         State status = getValue(STATUS);
         EntityFacade module = getValue(MODULE);
         EntityFacade path = getValue(PATH);
+        EntityFacade author = getValue(AUTHOR);
```
carldea, line 225: Instead of getting the author from the view model, couldn't you just get it from the `ViewProperties`.
Transactions would always write using the logged in user.

carldea:
```java
ConceptFacade loginUser = viewProperties.nodeView().editCoordinate().getAuthorForChanges();
```

dukke (approved):
Looks good to me, thanks Swaroop!

jdsmithsos (approved):
looks good

PR #676, opened 2025-08-25
feature/finished/IIA-2440-language-selection-in-parent-causes-child-override

jdsmithsos:
Jira tickets:

https://ikmdev.atlassian.net/browse/IIA-2440
https://ikmdev.atlassian.net/browse/IIA-2441

Summary of changes:  modified ObservableViewNoOverride.baseCoordinateChangedListenersRemoved() to set the value of existing LanguageCoordinates, rather than create new LanguageCoordinates objects

Before change:  Changing the Parent language causes child view coordinates to go into Overridden mode

Select Journal view coordinate language to Spanish then back to English.
Child goes into Overridden mode.

image

After change:  Child view coordinates no longer go into Overridden mode

image


jdsmithsos:
@swaroopsalvi There is not a parentID method available in LanguageCoordinate, which is why I am using the UUID

PR #677, opened 2025-08-26
Feature/finished/iia 2686 allow user to view unsupported field datatypes

carldea:
IIA-2686 Allows user to summon semantic (GeneralEditing Window) with field datatypes unsupported. Prior the user received an exception. Also some code to prevent a user from creating a semantic as an uncommitted record.

If a user attempts to create a semantic of a pattern with unsupported fields. It will prevent the user from continuing.


dholubek (approved):
looks good

PR #679, opened 2025-08-26
Feature/finished/iia 2698 unable to create semantic

carldea:
Fixed the class cast exception when creating a new Semantic.


dholubek (approved):
looks good

PR #681, opened 2025-08-26
Eclipse-Collections-13.0.0 Update

OmOmofonmwan:
- Updated Eclipse-Collections references to 13.0.0
- tinkar-core updated to 1.122.0
- tinkar-composer updated to 1.10.0
- Commented reference to coordinate since it is now located inside the entity module in tinkar-core


carldea (approved):
Looks good to me!
If tested we can merge.

PR #685, opened 2025-08-27
Feature/finished/iia 2603 identifier not being completely shown

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2603

Summary of changes:  

- Modified PublicIDControlSkin to resize the TextField to show all of the text within it
- Set the preferredWidth to the actual text width, plus the padding

Before changes:  The identifier text is truncated

image

After changes:

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/skin/PublicIDControlSkin.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/skin/PublicIDControlSkin.java
@@ -63,9 +63,6 @@ public class PublicIDControlSkin extends SkinBase<PublicIDControl> {
         publicIdTextField.setEditable(false);
         publicIdTextField.getStyleClass().addAll("public-id-textfield", "copyable-label");
 
-        // using setHgrow causes the TextField to grow to the full width of the HBox, which is not desirable
-//        HBox.setHgrow(publicIdTextField, Priority.ALWAYS);
-
         // the SVG graphic for the copy to clipboard icon
         var svgPath = new SVGPath();
         svgPath.setContent(SVGConstants.COPY_TO_CLIPBOARD_SVG_PATH);
@@ -77,9 +74,6 @@ public class PublicIDControlSkin extends SkinBase<PublicIDControl> {
         Tooltip.install(copyToClipboardButton, new Tooltip("Copy to Clipboard"));
 
         publicIdHBox.getStyleClass().add("public-id-box");
-
-        // using setHgrow causes the TextField to grow to the full width of the HBox, which is not desirable
-//        HBox.setHgrow(publicIdHBox, Priority.ALWAYS);
 
         // the publicIdHBox contains the publicIdLabel and the copyToClipboardButton.
         // Both controls need to be in a single HBox to be able to show and hide the


carldea:
@jdsmithsos  can we have @dukke  weigh in on this?

I don't think we should be creating a new Text object and setting width. There should be a way to `setPreferredWidth(-1)` or `setMaxWidth()` or something in the CSS to use computed width or available space horizontally.

jdsmithsos:
Changing the TextField to a Label would easily solve this, because the Label will be sized to the text, or be truncated with ellipsis, automatically.  But TextField is required so a copy can occur from the TextField.  But now with the Copy to clipboard button, I don't think that the TextField is necessary anymore.

dukke (changes requested):
> @jdsmithsos can we have @dukke weigh in on this?
> 
> I don't think we should be creating a new Text object and setting width. There should be a way to `setPreferredWidth(-1)` or `setMaxWidth()` or something in the CSS to use computed width or available space horizontally.

@carldea is making a good point.
@jdsmithsos custom control's skin has a computePrefWidth/Height, computeMinWidth/Height and computeMaxWidth/Height. You can override those methods in the skin and return the pref, min or max (depdning on the method you're overriding) that makes sense for the custom control.

That way the parent container of the control can adjust the child custom control size depending on what the control return as its min/pref/max width or height. This will make it dynamically adjust.

This would be the prefered way IMO.

As @carldea  is saying you can also leverage calling prefWidth(-1), minWidth(-1) etc on each control inside the custom control to get the preferred or min or max, width or height for each individual control that is part of the custom control.

dukke:
@jdsmithsos @carldea 
One other thing I'd add. The best possible solution for this would be to use a TextFlow. Problem is it might take more time to implement and that's why I initially suggested continuing to use the TextField.
One other option is to keep using a TextField solution but if there's no way for the Textfield to grow automatically to fit the text that's in it we would need to create a custom control that has a TextField and does that.

### Solution using a TextFlow:
**Cons**: my guess is that it'll take more time to implement
**Pros**: it's the best solution. Makes more sense since we only want selectable text (not a textfield).
**Details of the solution**: Contrary to Label, TextFlow has methods for you to check where the mouse was clicked (hitTest(Point2D), you can use that information and do the selection manually yourself (you would probably have to render the selection box yourself).

### Solution using a custom growing TextField:
**Cons**: you'd still  be using a textfield but you only want to show text
**Pros**: my guess is that this would be quicker to implement.
**Details of the solution**:
You would create a custom control class that only has a TextField. To know what the size of the text inside the TextField is you can use the lookup() method to get the Text node inside of the control and then calculate what's it's prefered size. You might need to have a custom layoutChildren method inside the Skin where you do the layout yourself based on the former infomation. You might also need to override computePrefWidth in the Skin, so that you return a default appropriate width that is based on the size of the Text node inside the TextField.
Then you could use this custom TextField that grows inside the PublicIDControl.

Thanks!

kview/src/main/java/dev/ikm/komet/kview/controls/skin/PublicIDControlSkin.java
```diff
@@ -107,8 +115,19 @@ public PublicIDControlSkin(PublicIDControl control) {
...
+
+            Platform.runLater(() -> {
+                // set the preferredWidth of the TextField to completely show the identifier text
+
+                Text text = new Text(identifier); // Create a temporary Text node with the new text
+                text.setFont(publicIdTextField.getFont()); // Set the same font as the TextField
+                double width = text.getLayoutBounds().getWidth() +
+                        publicIdTextField.getPadding().getLeft() +
+                        publicIdTextField.getPadding().getRight() + 2d; // Add padding and a small buffer
+                publicIdTextField.setPrefWidth(width);
```
carldea, line 129: Imo, We should handle this another way. 
publicIdTextField should have css applied and examine the outer parent. I can't remember but I think it is a gridpane cell. could be a colspan.

kview/src/main/resources/dev/ikm/komet/kview/controls/public-id-control.css
```diff
@@ -27,7 +27,10 @@
 }
 
 .public-id > .public-id-box > .public-id-textfield {
-    -fx-pref-width: 136;
+    /* The preferred width, set to display the first public ID in the TextField.
+       This value will change based on the -fx-font-size setting above.
+       The value used must work on the Concept, Pattern, and Semantic windows. */
+    -fx-pref-width: 195;
```
carldea, line 33: Can you try use the computed size? -1 for pref-width.

carldea (approved):
Looks good. I think this is sufficient for now.

carldea (approved):

PR #687, opened 2025-08-27
feature/finished/IIA-2720: Can't load files in Scene Builder that have the new StampViewControl

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2720

-------------

image


carldea (approved):
Looks good to me!
I guess this gets called during loading.

dukke:
> Looks good to me! I guess this gets called during loading.

yes! The listener gets set right during the construction of the skin. Then any mouse clicks will trigger the method to be called, if there is no parent window a NPE is thrown. 
Unfortunately, there's no "designer mode" for controls. I remember a C# library where you could check whether the control was being displayed in designer mode or not. If you do anything on the skin that's not supported scene builder won't be able to open it.

PR #689, opened 2025-08-28
Add Language Coordinates to FilterOptions control

jperedadnr:
This PR adds the LangFilterTitledPane control to display one set of Language Coordinates (for a given language). 

The FilterOptionsControl is basically formed by a root Accordion with the set of FilterTitledPane controls for the regular options (including a DateFilterTitledPane for the Time option), and a regular TitledPane that holds a second Accordion control with one or more LangFilterTitledPane controls.

FilterOptions model has been refactored to have the regular options (under MainCoordinates) plus a list of LanguageCoordinates (with at least one item).

Both NextGen Search and ConceptNavigator controllers have been refactored, extracting the loadFilterOptions method to a convenient FilterOptionsUtils class.

Note that the order in which the default FilterOptions model is filled up is really important. For that, see 
FilterOptionsPopupSkin::setupDefaultFilterOptions. Here the available options are set from either the resource bundle, or from the dataset (via Navigator), and then default/selected options are inherited from the ViewProperties (which in some cases are already set in the FilterOptions directly).

image

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/FilterOptions.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/FilterOptions.java
@@ -3,7 +3,6 @@ package dev.ikm.komet.kview.controls;
 import java.io.Serial;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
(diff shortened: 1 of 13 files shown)


dholubek (approved):
looks really good.  thank you for the code clean up as well.

PR #690, opened 2025-08-28
feature/finished/IIA-2729-dont-allow-deselection-of-include-or-exclude-module-wildcard

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2729

Summary of changes:

- In ViewMenuTask
    - Added check for selected for the Include wildcard CheckMenuItem
    - Added check for selected for the Exclude wildcard CheckMenuItem
    - If not checked, just rechecks the CheckMenuItem
    - If checked, then clears the Set property
- Set and ListPropertyWithOverride.clear() - put the final clear() call into an else, because no reason to clear a newly created empty Set/List

Before changes:  The wildcard menu item can be deselected

After changes:  The wildcard menu item CANNOT be deselected


carldea (approved):
Looks good to me!

PR #696, opened 2025-08-29
IIA-2621 Incorrect Field number display in Pattern

bhharsh13:
Fixed the field number display after updating the fields - https://ikmdev.atlassian.net/browse/IIA-2621

PFA for reference for the validations done on local.

**Before Change:**
image-20250819-141648

**After Change:**

Field 2 before update
Field 2 edit
Field 2 sequence updated correctly

### DIFF
--- a/kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
+++ b/kview/src/main/resources/dev/ikm/komet/kview/mvvm/view/kview.css
@@ -4163,7 +4163,7 @@ Styling a context menu for kview ui/ux controls
     -fx-font-weight: 500;
 }
 .card-info-container .validation-counts {
-    -fx-text-fill: -Accent-color-red;
+    -fx-text-fill: -Grey-12;
     -fx-font-family: Noto Sans;
     -fx-font-size: 25px;
     -fx-font-style: normal;


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
```diff
@@ -232,7 +232,7 @@ private ViewProperties getViewProperties() {
     private void clearView(ActionEvent actionEvent) {
         //Reset the field order selection to the last value
         IntegerProperty totalFields = patternFieldsViewModel.getProperty(TOTAL_EXISTING_FIELDS);
-        patternFieldsViewModel.setPropertyValue(FIELD_ORDER, (totalFields.get()+1));
+        patternFieldsViewModel.setPropertyValue(FIELD_ORDER, (totalFields.get()));
```
swaroopsalvi, line 235: This is not the fix, I think the FIELD_ORDER dropdown is for populating the edit box. If you do this then the user will not be able to select the next field order.

swaroopsalvi (changes requested):
The change should be in patternDetails controller.

dholubek (changes requested):
this isn't working... do create pattern.  add one field, add a second field (they will be 1,2) go to add a third field but put order=1 and you will get this (two fields 1's):
image

bhharsh13:
Fixed the Field number display issue. Validated the changes on local and looks good. @swaroopsalvi / @dholubek , please help to approve the PR. PFA for reference.
**Before Adding New Field:**
Before adding Field 5
**After Clicking DONE** - Field number is updated correctly.
Field 5 added
**Change the Field number while editing existing fields**
Change the field order 5 to 1
**Field Order is updating correctly:**
Field order updated

dholubek (approved):
looks good, Bhavani confirmed she could add a stamp at place 1 in front of 2 and 3.

PR #697, opened 2025-08-29
IIA-2727 Remove Fake Error Alerts

bhharsh13:
Removed the Fake error alerts. "You have 25 errors" is replaced with “No New Messages” text. - https://ikmdev.atlassian.net/browse/IIA-2727

**Before Change:**
nonew messages

**After Change:**
Remove Fake Error Alerts

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
-        <version>1.55.0-SNAPSHOT</version>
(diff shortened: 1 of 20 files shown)


carldea (approved):
Looks good to me!

swaroopsalvi (changes requested):
I Think all the boxes need to be removed. You are seeing the concept on those Error cards that should to be removed as well.

cc: @carldea , @dukke @dholubek

dholubek (changes requested):
do we want it red and that large?  seems like red is more attention grabbing and more of an error color...

PR #699, opened 2025-09-01
JPRO-183 fix window resize issues around all propertie descriptions

vollbrecht-work:
This contains only the hotfix part of https://github.com/ikmdev/komet/pull/688.


carldea (approved):
Excellent work!

Responsive layout or custom layouts are the way to go. In the future refactoring many of these early mechanisms/strategies would be ideal.

PR #703, opened 2025-09-02
Add DateComboBox control to select stamps

jperedadnr:
Time FilterOption.Option now uses long timestamp (as string), and not just a local date (as a string), in order to pass along the stamp selected by the user with the new comboBox:

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/FilterOptionsUtils.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/FilterOptionsUtils.java
@@ -14,9 +14,7 @@ import dev.ikm.tinkar.entity.StampService;
 import dev.ikm.tinkar.terms.ConceptFacade;
 import org.eclipse.collections.api.list.primitive.ImmutableLongList;
 
-import java.text.ParseException;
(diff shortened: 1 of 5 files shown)


carldea:
@jperedadnr  I approved, but I will update branch, merge.

kview/src/main/java/dev/ikm/komet/kview/controls/FilterOptionsUtils.java
```diff
@@ -105,24 +101,22 @@ public static FilterOptions loadFilterOptions(ObservableCoordinate<ViewCoordinat
...
-            return -1;
+            return -1L;
         }
 
-        Date date;
         try {
-            date = SIMPLE_DATE_FORMAT.parse(time.selectedOptions().getFirst());
-        } catch (ParseException e) {
-            throw new RuntimeException(e);
+            return Long.parseLong(time.selectedOptions().getFirst());
```
dholubek, line 108: @jperedadnr is there a situation where this could be the string "Latest" and the parseLong will throw an exception?

dholubek (changes requested):
I just had one question

PR #707, opened 2025-09-03
feature/finished/IIA-2739-2741-view-coord-menu-catch-viewcalc-exception

jdsmithsos:
Jira tickets:

https://ikmdev.atlassian.net/browse/IIA-2739
https://ikmdev.atlassian.net/browse/IIA-2741

Summary of changes:

- refactored calls to the viewCalculator to catch Exception thrown, but still provide a String value to the menu is rendered and working

Before changes:  view coordinate menu went into overridden mode, but was not rendered correctly and not working

After changes:  view coordinate menu is now rendered and works correctly


carldea (approved):
Looks good to me.

Please have @swaroopsalvi look at this too. Not sure getPreferred... does what we think it does with respect to language preferences. I think it is fine for now. :-).

PR #719, opened 2025-09-05
feature/finished/IIA-2564-fix-for-edit-fqn

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2564

Summary of changes:  moved the call to setConceptAndPopulateForm() outside of the check for the pane not being the editFqnPane, so the editFullyQualifiedNameController is always set with the publicId

Before change:

Edit the FQN, change the case significance to Case Sensitive, and Submit

image

Click on the FQN to edit, but the properties are set to blank or default values (not being set)

image

After change:

Edit the FQN, change the case significance to Case Sensitive, and Submit

image

Click on the FQN to edit, and now the properties are populated

image


bhharsh13 (approved):
Looks good to me

PR #721, opened 2025-09-05
JPRO-cleanup rename DetailsController and related methods info ConceptController

vollbrecht-work:
This touches only kview and ignores the Classic Komet "details" module


vollbrecht-work:
ah that was a wired merge conflict. Between locally, my fork and origin/main the module-info changes got lost... should work now

PR #723, opened 2025-09-05
IIA-2711 identifier name identifier icon display fix

bhharsh13:
Current Implementation was displaying Identifier number and Identicon Image for the existing reference component. When reference component was null , the condition was not handled. I have implemented the fix for this. Validated on my local and it looks good. PFA for reference.
https://ikmdev.atlassian.net/browse/IIA-2711

**Before Change:**
Identifier Identicon not updated

**After Change:**
Identifier Identicon updated
Referenced component


dholubek:
closing this PR as we will create a new one

PR #724, opened 2025-09-05
Show orange dot when filter options has overrides

jperedadnr:
This PR syncs the filter button orange dot for both Next Gen Search and Navigator Control Search with the overrides of the filter options.


carldea (approved):
Looks good to me!
I will approve and merge.

PR #725, opened 2025-09-05
Feature/finished/IIA-2820: When user tries to edit the pattern, the publish button is not getting enabled

dukke:
Fixes: https://ikmdev.atlassian.net/browse/IIA-2820

----------------------

Fixes when user tries to edit the pattern, the publish button is not getting enabled


carldea (approved):
I believe this looks okay. I will approve, can you please merge at your convenience?


**Technical Notes:**
I'm noticing nesting view models into view models. for example:
```java
patternViewModel.getProperty(STAMP_VIEW_MODEL).bind(propertiesController.stampFormViewModelProperty());
```

In the early stages when incorporating `ViewModel`s I nested a view model (`StampEditViewModel`) into another view model (`ConceptViewModel`). 

I believe it worked (sort of) but it did not feel right. It seemed to add yet another layer of indirection for no reason. 

In the future we should take a look at the classes `dev.ikm.tinkar.component.Stamp` or `dev.ikm.tinkar.entity.StampRecord` as it relates to viewing and creating stamps.

PR #726, opened 2025-09-08
JPRO-195 change ObjectBinding computeValue to not capitalize the value of KLReadOnlyDataTypeControl

vollbrecht-work:
I am not sure why this was added that way but it seams wrong to me that this was done in the Skin and not to the data itself?

The related issue suggested that its a problem. The only question that remain is if in other places we have wrong data and henceforth capitalization would be needed.


gsirojev:
@carldea @dukke it was TE feedback/[requirement ](https://ikmdev.atlassian.net/issues/?jql=textfields%20~%20%22capitalization%2A%22&selectedIssue=IIA-2730&wildcardFlag=true). They noticed that when editing a semantic field, values that are a string data type are being automatically capitalized when they should not be. and @vollbrecht-work worked on that and submitted PR for that

PR #727, opened 2025-09-08
JPRO-cleanup Move Concept specific controllers out of properties into concept

vollbrecht-work:
This is a first step to more clearly represent the **current** hirachical structure.

This will hopefully in the future make it simpler to actually create "shared" properties classes.

This PR does not move out History / Hierarchy Controllers. Since they currently are not exclusivly used by the ConceptController, but also still part of kview.lidr.mvvm.view.properties.


dholubek:
@vollbrecht-work can you resolve the merge conflicts?  can you test create concept and summon a concept from the concept navigator to verify that everything is working?  I pulled down the branch and got an exception when choosing create concept.

dukke (changes requested):
Adding to @dholubek point I would test Concept creation and editing an existing Concept before merging, if possible.

carldea:
I'm wondering if we should redo this PR, because it's out of date. @vollbrecht-work Let me know your thoughts.

PR #729, opened 2025-09-08
JPRO-192 Rework fxml to be more inline with concept details view.

vollbrecht-work:
made the design similar to the figma base design by aligning the semantic description directly under the sematic title

implemented correct wraping behaviour and expanding behaviour for the details

changed the stampview part from beeing inside a anchorpane to a stackpane allowing for simpler integration of upcoming changes already applied in concept view

still continue to providing a identifier field, while this element is missing in the figma for Pattern Semantics

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/controls/skin/PublicIDControlSkin.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/controls/skin/PublicIDControlSkin.java
@@ -111,17 +111,6 @@ public class PublicIDControlSkin extends SkinBase<PublicIDControl> {
             identifier = publicId;
             publicIdTooltip.setText(identifier);
             publicIdTextField.setText(identifier);
-
-            Platform.runLater(() -> {
-                // set the preferredWidth of the TextField to completely show the identifier text
-
-                Text text = new Text(identifier); // Create a temporary Text node with the new text
-                text.setFont(publicIdTextField.getFont()); // Set the same font as the TextField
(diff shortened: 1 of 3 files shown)


carldea (approved):
Great work. I will merge once you have resolved the merge conflict.

vollbrecht-work:
yeah i noticed we have another problem here with the IDENTIFIER, with the recent changes in https://github.com/ikmdev/komet/commit/acc0590f058dc8a87adc530f68333b056cce8901#diff-ef07c5ed0b2ca884d8fd1d0fe2b2887c3f37533003615756cd1032f117b25609R115 and how its embedded in the view overall correct inner with is not correctly provided. E:g its possible to shrink the window here beneath the value.

carldea:
> a video in the jira ticket JPRO-192 since its to big for here.

Yep, all stamp popups using the `...` (3 dots button), are being phased out. All windows (not including LIDR) will be using the new way to enter Stamp information. Rule of thumb. When creating things such as a concept, pattern or semantic the user must enter a valid Stamp before proceeding. Event though the UI allows them to randomly go to sections, the user could experience an exception because of the stamp requirement. 

cc/ @jperedadnr

carldea:
@vollbrecht-work Fredrick update your branch again. ;-). I can merge if you don't have any changes. let me know. I will give you some time to respond before I merge.

vollbrecht-work:
@carldea this should at least make look it good for the genedit window

PR #730, opened 2025-09-08
Feature/finished/iia 2831 classic parent menu

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2831

classic menu has the "logo banner" plus the File menu above it now:
image

journal on windows desktop maintains the menu:
image

and tested with JPro for the classic menu:
image

and JPro's menu on the journal view:
image

### DIFF
--- a/application/src/main/java/dev/ikm/komet/app/AppPages.java
+++ b/application/src/main/java/dev/ikm/komet/app/AppPages.java
@@ -259,7 +259,7 @@ public class AppPages {
         journalStage.setScene(sourceScene);
 
         if (!IS_MAC) {
-            app.appMenu.generateMsWindowsMenu(journalBorderPane, journalStage, null);
+            app.appMenu.generateMsWindowsMenu(journalBorderPane, journalStage);
         }
 
         // load journal specific window settings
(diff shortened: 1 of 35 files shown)


application/src/main/java/dev/ikm/komet/app/AppPages.java
```diff
@@ -260,7 +259,7 @@ void launchJournalViewPage(PrefX journalWindowSettings, ConceptFacade loggedInUs
         journalStage.setScene(sourceScene);
 
         if (!IS_MAC) {
-            app.appMenu.generateMsWindowsMenu(journalBorderPane, journalStage);
+            app.appMenu.generateMsWindowsMenu(journalBorderPane, journalStage, null);
```
carldea, line 262: Can you have two overloaded methods? Keeping the original for those callers.
```java
  app.appMenu.generateMsWindowsMenu(journalBorderPane, journalStage);
```

The existing libraries would be documented explaining the new third argument. 
```java
  
  void generateMsWindowsMenu(BorderPane kometRoot, Stage stage) {
         this.generateMsWindowsMenu(journalBorderPane, journalStage, null);
  }
  
  void generateMsWindowsMenu(BorderPane kometRoot, Stage stage, VBox topBarVBox) {
      ...
  }
```
Can you also document above the methods?

**Tech notes:** 
Null logic or an argument denoting 'not present' is not good practice. The user of the api or reader of the code at glance would know when or why the third parameter is null and/or what kind of object to pass into.

dholubek: updated

dholubek:
> Can we also ask QA to test the PR before we merge? This must work on the web browser, Windows, and MacOS as an installed application. In JavaFX menus placed into a VBox may not be the desired layout behavior. menus usually span the width of the application at the top. But it appears correctly in the screen shots above.

@carldea I created a subtask for this to be tested: https://ikmdev.atlassian.net/browse/IIA-2841

carldea (approved):
Looks good to me!

PR #733, opened 2025-09-09
Feature/finished/iia 2796 child view coordinate menu item text fix

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2796

Summary of changes:

- ViewMenuModel
    - It was discovered that ViewMenuModel used the ViewCalculator differently than the JournalController
    - Modified ViewMenuModel to not create a new ViewCalculator when a view coordinate change event occurs
    - Refactored ViewMenuModel constructors to remove duplicate code
- ViewMenuTask - removed unnecessary try catch, which is coded within the getPreferredDescriptionStringOrNid() method and not necessary in the getNameAndValueString() method
- ObservableViewBase - added new constant DEFAULT_VIEW_NAME rather than hard-coding the "View" string in the constructor

Before changes:  XML text was displayed in the child view coordinate menus

In a child view coordinate menu select INACTIVE or WITHDRAWN

image

The view coordinate menu has XML text

image

After changes:  Correct text is displayed in the child view coordinate menus

In a child view coordinate menu select INACTIVE or WITHDRAWN

The view coordinate menu no longer has XML text

image

### DIFF
--- a/framework/src/main/java/dev/ikm/komet/framework/view/ObservableViewBase.java
+++ b/framework/src/main/java/dev/ikm/komet/framework/view/ObservableViewBase.java
@@ -50,6 +50,9 @@ public abstract class ObservableViewBase
     protected final ObservableLogicCoordinateBase logicCoordinateObservable;
 
     protected final ObservableEditCoordinateBase editCoordinateObservable;
+
+    public static final String DEFAULT_VIEW_NAME = "View";
+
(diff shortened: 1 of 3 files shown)


carldea (approved):
Looks good to me.
Nice to clean up things too (for readability).

PR #735, opened 2025-09-09
Feature/finished/iia 2824 populate all fq ns for a concept when there is more than one fqn

swaroopsalvi:
https://ikmdev.atlassian.net/browse/IIA-2824
 https://ikmdev.atlassian.net/browse/IIA-2825

image

image

image

Before changing coordinates:
image

=================================================

https://ikmdev.atlassian.net/browse/IIA-2835
https://ikmdev.atlassian.net/browse/IIA-2836

Before Change coordinate

image

Change Preferences:
image

image

Change Language:
image
image


carldea (approved):
Excellent work!

**Tech note:**
I know the description type's combobox is populated with entries and it's read-only it's okay for now since the early designs allowed the user to choose whether it was a fqn or regular name. 

A new feature would be to allow the user to edit the description type of 'DEFINITION'. Only way to add or edit a definition is through the GenEditWindow (Semantic create/edit). 

A bigger effort would be to consolidate all the description add/edit panels which relates to this PR: https://github.com/ikmdev/komet/pull/688
cc/ @vollbrecht-work

PR #737, opened 2025-09-10
Feature/finished/iia 2456 journal numbering

dholubek:
After closing and reopening journal windows, the window title and the journal title were getting out of sync.  This fixes that.
Previously the AtomicCounter singleton would get the latest number.  This is fine if you only just create new windows and don't close and reopen.  But it will get the highest number when you reopen say 5 of 10 and will say '10' in the blue portion of the title.  The atomic counter is really for the landing page to increment to the latest.  If we store the title in the viewModel then we are always synced to the correct title.

image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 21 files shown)


carldea (approved):
Great work @dholubek !
Thanks for testing many permutations.

dholubek:
> Great work @dholubek ! Thanks for testing many permutations.

np

PR #740, opened 2025-09-10
Feature/iia 2424 dnd semantic in top comp search

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2424

drag
image

and drop
image


carldea:
@dholubek  I just updated your branch.

PR #741, opened 2025-09-10
Fix for bug IIA-2836 Definition under concept title and IIA-2782

swaroopsalvi:
https://ikmdev.atlassian.net/browse/IIA-2782
https://ikmdev.atlassian.net/browse/IIA-2836

Logged In User:
image

Before editing AXIOM:
image

After changing Axiom:
image

=============================
After changed Preferences:
image

image

Change language to Spanish.
image

image


carldea (approved):
Excellent work!

Nice that you tracked down all the places with the hardcoded user.

PR #743, opened 2025-09-10
Feature/finished/iia 2605 composite identifier custom control

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2605

Summary of changes:

- Created a new custom control PublicIDListControl
- The PublicIDListControlSkin uses the existing custom control PublicIDControl to get the Copy to clipboard functionality
- Modified PublicIDControl to be able to hide the identifier label, which is not needed with the new identifier list provided in the new PublidIDListControl
- Modified the Concept window to use the new PublicIDListControl
- Modified the Concept window fxml layout to move the Description down to the Descriptions pane

Before change:  The IDENTIFIER is displayed by the PublicIDControl in the Concept window

image

After change:  A list of identifiers is now displayed.  The screenshot below shows a vertical scroll bar, which is available if the VBox height exceeds the available height, which is more than 2 identifiers.

image

The copy and pasted identifiers from the new PublicIDListControl:

image


vollbrecht-work:
how many new vertical identifiers do we expect here? If they are already there (e.g part of the dataset) when the concept is opened, then it should be no problem to display them without a scrollbar. Is it planed here to have a custom number of them depending on the concept or are they staying the same across all current concepts?

PR #755, opened 2025-09-15
Feature/finished/IIA-2874: [Pattern Stamp Editing] New version of Stamp Edit mode for Pattern Window

dukke:
Fixes: 

1. https://ikmdev.atlassian.net/browse/IIA-2874
2. https://ikmdev.atlassian.net/browse/IIA-2722

-------------------------

This PR implements a second version/iteration of Stamp "editing" for Patterns based on feedback that was given for the first iteration.
The user will now need to press confirm on the Stamp form and then hit publish for a new Pattern version with the new Stamp information to be published.

Also fixes a bug where the submit/confirm button would still remain enabled after submitting a new stamp version and the stamp values on the form would still be the same.

image


dukke:
> Looks good to me.
> 
> **Tech note:** I am seeing a pattern emerge where view model properties will own a `public enum` containing properties named simply `Properties` (as opposed to `StampProperties` which seems redundant). I am beginning to favor this. As java allow public enums and public class to be in the same file. I prefer static imports so what's left is the property name itself. The downside maybe if a class contains two different view models that share the same property name , but in that case you'd fully qualify the name.

@carldea I think there's a fair point. I'll change that in the next PR. Perhaps we could also change "StampType" enum name to just "Type"

PR #756, opened 2025-09-15
JPRO-169-171 Allow statechanges of Property-{Concept/Pattern/Semantic}-Windows to persiste

vollbrecht-work:
Since this commit touches the preference backing system, it will only work on journalwindows that once got its panes recreated.

- fix bug that did not allow custom prop's to persist if there type was none boolean

- fix small fxml bug in Pattern/PatternSematics FXML around CommentsButton

- property patterns are currently saved with custom STRINGS representing there state In a later PR this can be more generally hocked into ModelView state of Windows.

Notice that this commit allows for statechanges to persist through exiting and reopening the respective Journal. On the Pattern window for example closing and opening the property tab will **not** remember that state as this is handled internally in the window itself. Once the situation is rectified there the solution can directly be hooked up into this solution. 

Also notice that any persistent state changes are only seen, if a window itself tells about that changes. This is currently not correctly done if nothing visually changes on JournalClose. E.g to workaround that case, move the window a tiny amount. (Such action immediately save the current state)  Rectifying that is not part of the linked issues.


carldea (approved):
This looks good for now. I will approve and merge.

**Tech Note:**
For the sake of time. I will accept the preferences code to serialize/deserialize name value pairs. I had forgotten about KometPreferences has APIs that support different datatypes to persist and retrieve. If time permits (which I doubt it) would could refactor to make use of that facility in another ticket.

PR #757, opened 2025-09-15
IIA-2879: NullPointerException while updating Pattern Meaning

bhharsh13:
Fixed the NullPointerException while updating Pattern Meaning - https://ikmdev.atlassian.net/browse/IIA-2879.
Please help to review the PR.

**Before Change:**
Before Fix

**After Change:**
image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
@@ -128,7 +128,8 @@ public class PatternFieldsController {
             if (meaningObject != null) {
                 ConceptEntity conceptEntity = Entity.getFast((EntityFacade) meaningObject);
                 if (conceptEntity != null && viewProperties != null) {
-                    displayNameTextField.setText(getDescriptionTextWithFallbackOrNid(conceptEntity, viewProperties));
+                    LanguageCalculator languageCalculator = viewProperties.calculator().languageCalculator();
+                    displayNameTextField.setText(languageCalculator.getDescriptionTextOrNid(conceptEntity.nid()));
                 } else {
                     displayNameTextField.setText("");
                     LOG.warn("conceptEntity is null.");


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
```diff
@@ -121,8 +126,9 @@ private void initialize() {
         patternFieldsViewModel.getProperty(MEANING_ENTITY).subscribe(meaningObject -> {
             if (meaningObject != null) {
                 ConceptEntity conceptEntity = Entity.getFast((EntityFacade) meaningObject);
-                if (conceptEntity != null) {
-                    displayNameTextField.setText(conceptEntity.description());
+                if (conceptEntity != null && viewProperties != null) {
+                    LanguageCalculator languageCalculator = viewProperties.calculator().languageCalculator();
+                    displayNameTextField.setText(languageCalculator.getFullyQualifiedDescriptionTextWithFallbackOrNid(conceptEntity.nid()));
```
carldea, line 131: Question: Shouldn't we be calling `getDescriptionTextWithFallbackOrNid()`? 
It appears you are essentially forcing to display a fully qualified description. 
What if the user may change the view coordinates that would intern change the display. Like the following:
Screenshot 2025-09-15 at 12 07 30 PM

Also I'm not sure if `displayNameTextField` is the correct field you are to set. Is this the meaning or the name of the Pattern?

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
```diff
@@ -118,7 +118,7 @@ private void initialize() {
             if (meaningObject != null) {
                 ConceptEntity conceptEntity = Entity.getFast((EntityFacade) meaningObject);
                 if (conceptEntity != null) {
-                    displayNameTextField.setText(conceptEntity.description());
+                    displayNameTextField.setText(viewProperties.calculator().getPreferredDescriptionTextWithFallbackOrNid(conceptEntity.nid()));
```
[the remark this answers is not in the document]
carldea: @bhharsh13 Swaroop is correct in using getDescritionTextFallBackNid()

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/pattern/PatternFieldsController.java
```diff
@@ -121,8 +127,8 @@ private void initialize() {
         patternFieldsViewModel.getProperty(MEANING_ENTITY).subscribe(meaningObject -> {
             if (meaningObject != null) {
                 ConceptEntity conceptEntity = Entity.getFast((EntityFacade) meaningObject);
-                if (conceptEntity != null) {
-                    displayNameTextField.setText(conceptEntity.description());
+                if (conceptEntity != null && viewProperties != null) {
+                    displayNameTextField.setText(getDescriptionTextWithFallbackOrNid(conceptEntity, viewProperties));
```
carldea, line 131: Can you use the view calculator's `getDescriptionTextWithFallbackOrNid()`. I would not rely on `ViewCalculatorUtils.java`. Underneath the covers is not calling the same method. 

**Tech Note:**
Ui developers should use the view calculator's methods often. Helper methods are good for situations to cut down on verbose code, however helpers are newer and not tested. For example: 
```java
  String text = ViewCalculatorUtils.getDescriptionTextWithFallbackOrNid(...); 
  // actually calls getPreferredDescriptionTextWithFallbackOrNid
```
Screenshot 2025-09-15 at 1 03 24 PM

You want to call `viewProperties.calculator().getDescriptionTextOrNid()`
I think we (Swaroop and I) said with fallback nid. it was off the top of our heads so we meant above.

PR #758, opened 2025-09-15
J pro 191

thorstenstueker:
This PR changes the behavior in case of NULL kids which are causing an error when syncing with GitHub.


carldea:
@thorstenstueker  I have recently updated this branch. I hope you got a chance to address the comments above.

thorstenstueker:
Yes I have removed the not needed import and module requirement from my sources.I have got another error in the same class which is not connected to this one.

PR #767, opened 2025-09-16
Feature/finished/IIA-2873: Allow properties panel to open automatically when user creates a new Semantic AND IIA-2843

dukke:
Fixes:
1. https://ikmdev.atlassian.net/browse/IIA-2873
2. https://ikmdev.atlassian.net/browse/IIA-2843

------------------------

Fixes:
1 - Semantic creation - NPE on adding a ref component
2 - Allow properties panel to open automatically when user creates a new Semantic


carldea (approved):
Looks good to me!

PR #768, opened 2025-09-16
JPRO-200 allow for state changes in FilterMenuPopup to persist

vollbrecht-work:
Implementing a ViewModel to detach some of the pathMap state into the ViewModel

This can further be utilized to also detach the TimelineController from more state, like its done in the new FilterMenuController.

This PR also added the bonus feature that any changes in the FilterMenu are now instantly updated in the Timeline itself.


carldea:
@vollbrecht-work  Please update the branch.

carldea (approved):

PR #774, opened 2025-09-18
Bind nodeView with FilterOptions

jperedadnr:
Add bindings between nodeView and FilterOptionsPopup


carldea (approved):
Excellent work! 

Thank you for pointing out possible side effects if another area of code binding or changing the child view coordinate properties.

PR #777, opened 2025-09-18
Feature/finished/iia 2909 identifier copy to clipboard button

jdsmithsos:
Jira ticket:  https://ikmdev.atlassian.net/browse/IIA-2909

Summary of changes:

- Simplified PublicIDControl, now that all three windows Concept, Pattern and Semantic are using the same PublicIDListControl for identifier display
- With permission from Jieun, changed the TextField to a Label in PublicIDControl

Before changes:  The Copy to clipboard is not visible when the width of the Concept window is reduced

image

After changes:  The Copy to clipboard button is visible

image


bhharsh13 (approved):
looks good to me.

PR #778, opened 2025-09-18
Feature/finished/iia 2929 top comp names respond coords

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2929

Changing the preferred from OT,FQN to FQN,OT.  Classic Komet uses this line of code to get the top component name and it works there so we can do the same to make it work in next gen:

https://github.com/ikmdev/komet/blob/5c687ada3070490c8a0a30e94a899990dfd87159/framework/src/main/java/dev/ikm/komet/framework/search/SearchPanelController.java#L248

Here is the FQN and OT:
image

default grabs OT first:
image

we can apply an override and pick the FQN first:
image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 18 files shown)


bhharsh13 (approved):
looks good to me

PR #779, opened 2025-09-18
IIA-2912 Semantic Title Not Updated Until Semantic Window Is Reopened

bhharsh13:
Fixed the Semantic Title display -https://ikmdev.atlassian.net/browse/IIA-2912

Summary of Changes:
Added a listener to refComponentProp, wrap the UI update inside Platform.runLater.

**Before Change:**
image-20250917-140527

**After Change:**
New Semantic title
After adding Reference Component

Semantic title updated

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
@@ -262,7 +262,7 @@ public class GenEditingDetailsController {
         // Setup Properties Bump out view.
         setupProperties();
         //Populate the Title Pattern meaning purpose
-        setupSemanticForPatternInfo();
+        updateSemanticForPatternInfo();
         setupFilterCoordinatesMenu();
 
         //Populate readonly reference component.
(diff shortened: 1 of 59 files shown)


jdsmithsos (commented):
I just made the same kind of fix in Pattern window, but for the Identifier not being set after creation.  What I did was check the edit mode change to EDIT, then set the identifier.
See my PR https://github.com/ikmdev/komet/pull/770.

Does your change follow the same pattern that Pattern and Concept windows use to set the title after creation?

(Honestly I'm not sure what Concept and Semantic do to set the Identifier after creation, I didn't even look, but probably should now since I brought this up)

bhharsh13:
Addressed the code review comments. 
image

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
```diff
@@ -629,6 +629,7 @@ private void setupReferenceComponentUI() {
                 case ConceptEntity ignored -> {
                     refType = "Concept";
                     description = refComponent2.description();
+                    updateSemanticForPatternInfo();
```
swaroopsalvi, line 632: move this call to the end for the function after the 

`` referenceComponent.setTitle(refType);
            referenceComponent.setValue(componentItem);
``

swaroopsalvi (changes requested):

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
bhharsh13: Made the changes as suggested.
image

swaroopsalvi (approved):
looks good Approved.

PR #780, opened 2025-09-18
Feature/finished/iia 2906 semantic stamp incorrect last updated

swaroopsalvi:
This issue fixes
https://ikmdev.atlassian.net/browse/IIA-2906
https://ikmdev.atlassian.net/browse/IIA-2931
https://ikmdev.atlassian.net/browse/IIA-2932

Before Edit (AutoSave): Close window without hitting submit button.
image

After Submit (EDIT):
image

Create:

image

image

### DIFF
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
@@ -320,6 +320,7 @@ public class GenEditingDetailsController {
 
         if(genEditingViewModel.getPropertyValue(MODE).equals(CREATE)) {
             addReferenceButton.setDisable(!isConfirmedOrSubmitted);
+            stampViewControl.setDisable(isConfirmedOrSubmitted);
         }
     }
 
--- a/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
+++ b/kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
@@ -468,11 +468,8 @@ public class GenEditingDetailsController {
             semanticEntityVersionLatest = retrieveCommittedLatestVersion(observableSemanticSnapshot);
             //Set and Update STAMP values
             semanticEntityVersionLatest.ifPresent(semanticEntityVersion -> {
-                propertiesController.getStampFormViewModel().setPropertyValue(CURRENT_STAMP, semanticEntityVersion.stamp());
                 updateUIStamp(semanticEntityVersion.stamp().lastVersion());
             });
-            //Commenting out this method since the above line #470 does not updated the stampFormViewModel CURRENT_STAMP.
-      //      updateUIStamp(propertiesController.getStampFormViewModel());
         };
         subscriberList.add(refreshSubscriber);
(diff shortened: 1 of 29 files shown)


kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
```diff
@@ -460,22 +463,16 @@ private void setupSemanticDetails() {
...
             semanticEntityVersionLatest.ifPresent(semanticEntityVersion -> {
-//                StampEntity stampEntity = semanticEntityVersion.stamp();
-//                stampViewModel.setPropertyValue(STATUS, stampEntity.state())
-//                        .setPropertyValue(TIME, stampEntity.time())
-//                        .setPropertyValue(AUTHOR, stampEntity.author())
-//                        .setPropertyValue(MODULE, stampEntity.module())
-//                        .setPropertyValue(PATH, stampEntity.path())
-//                ;
-//                stampViewModel.save(true);
+                propertiesController.getStampFormViewModel().setPropertyValue(CURRENT_STAMP, semanticEntityVersion.stamp());
```
dukke, line 471: It's not clear in the view model classes but we shouldn't be updating the CURRENT_STAMP property outside of the view model. 
CURRENT_STAMP should be a read-only property that only gets updated by the view model itself (encapsulation).
If you call propertiesController.getStampFormViewModel().update(...) it should update all the view model properties appropriately.

kview/src/main/java/dev/ikm/komet/kview/mvvm/viewmodel/stamp/StampAddFormViewModelBase.java
```diff
@@ -60,10 +64,19 @@ protected void doUpdate(EntityFacade entity, UUID topic, ViewProperties viewProp
...
+        ObservableEntity observableEntity = ObservableEntity.get(entityFacade.nid());
+        ObservableEntitySnapshot observableEntitySnapshot = observableEntity.getSnapshot(viewProperties.calculator());
+        Latest<EntityVersion> latestEntityVersion = retrieveCommittedLatestVersion(observableEntitySnapshot);
+        latestEntityVersion
+                .ifPresentOrElse(latestVersion -> setPropertyValue(CURRENT_STAMP, latestEntityVersion.get().stamp())
+                    , () -> {
+                               EntityVersion latestVersion = viewProperties.calculator().latest(entityFacade).get();
+                               setPropertyValue(CURRENT_STAMP, latestVersion.stamp());
+                    }
+                );
```
dukke, line 76: Not sure if changing this code could break other Components for which this is also used and was working

kview/src/main/java/dev/ikm/komet/kview/mvvm/view/genediting/GenEditingDetailsController.java
```diff
@@ -557,6 +554,24 @@ private void populateSemanticDetails() {
...
+
+        EntityFacade author = stampEntityVersion.author();
+        stampViewControl.setAuthor(author == null? "" : getViewProperties().calculator().getDescriptionTextOrNid(author.nid()));
+
+        EntityFacade module = stampEntityVersion.module();
+        stampViewControl.setModule(getViewProperties().calculator().getDescriptionTextOrNid(module.nid()));
+        EntityFacade path = stampEntityVersion.path();
+        stampViewControl.setPath(getViewProperties().calculator().getDescriptionTextOrNid(path.nid()));
+
+    }
```
dukke, line 573: we should use the updateUIStamp method that already exists that takes a StampViewModel instance.

dukke (changes requested):

PR #799, opened 2025-09-23
Refresh text of options when navigator changes.

jperedadnr:
Changing the language to Spanish now refreshes the text of FilterOptions popup options:

image


carldea (approved):
Looks good to me!

PR #804, opened 2025-09-24
Feature/finished/iia 2910 remove added date from pattern definition and pattern fields

swaroopsalvi:
https://ikmdev.atlassian.net/browse/IIA-2910

image


swaroopsalvi:
> Rather than comment out, just delete those lines.

@jdsmithsos , Done.

jdsmithsos (approved):

PR #807, opened 2025-09-24
Use FxGet to retrieve Path options

jperedadnr:
And prevent exceptions from descendantsList.


carldea (approved):
Looks good. I think it's okay to keep the commented out line for the next team to help improve.

PR #808, opened 2025-09-25
JPRO-210 correctly save changed EntityFacade to preference system.

vollbrecht-work:
This does not only fix it for the Pattern window, but also for Concept and PatternSemantics/GenEditing in a general way.

The presented way of saving preferences on change in the inner ViewModel should be seen as the prefered way of updating state persistenly. In future work this code could be expaned to also handle saving preferences in a central clean way on ViewModel changes such as propertPanel state.


carldea (approved):
@vollbrecht-work 
Looks good to me!
Thank you Fredrick

PR #809, opened 2025-09-25
Select Any if all modules get deselected

jperedadnr:


carldea:
I've updated the branch and will merge.
Thank you @jperedadnr !

PR #811, opened 2025-09-25
Jpro refactored ConceptController

vollbrecht-work:
This is a nearly finished refactor of the ConceptController / ConceptViewModel. 

Main features:
- Streamlined state management ( all inner state is now contained in a ConceptViewModel) 
- Drastically reduces the previously duplicated code - for example such as in the old Semantic FormModels
- Create symmetry in the codebase with respect to the Pattern / PatternSemantic Window.


vollbrecht-work:
this PR would superseed #688 and #727

PR #821, opened 2025-09-25
Feature/finished/iia 2956 nid onname after reasoner

dholubek:
https://ikmdev.atlassian.net/browse/IIA-2957

We determined that the nid is expected since we are using the language calculator.  However, we can also add 'No version for view' in places where it is applicable.

image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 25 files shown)


carldea (approved):
Looks good to me. 
I think the user/developer has access to the Komet ID so no need to display the negative nid anyways.

dholubek:
> Looks good to me. I think the user/developer has access to the Komet ID so no need to display the negative nid anyways.

that is a good point

PR #822, opened 2025-09-25
IIA-2968 Komet ID alignment in Pattern window

bhharsh13:
Fixed Komet ID alignment in Pattern Window -https://ikmdev.atlassian.net/browse/IIA-2968

**Before Change:**
image-20250925-191725

**After Change:**
image

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>dev.ikm.komet</groupId>
         <artifactId>komet-parent</artifactId>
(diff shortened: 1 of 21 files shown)


carldea (approved):
Looks good to me. Thank you @bhharsh13 

I trust this is similar to the Concept Window based on column index/span. 

**Tech Note:**
  a good test would be to have many ids, such as komet id, snomed id, rxnorm id. That way it could show the custom control and it's copy and paste button.

PR #824, opened 2025-09-26
JPRO-206 fix description cutout on concept window rerender

vollbrecht-work:
This bug service through a race between minHeight and prefHeight calculations, where under rerender conditions a scenario is arived where the minHeight is bigger then the prefHeight when the prefHeight Skin ovverride is executed, confusing the render enginge.

Since in layouting this particular Node we only care about prefHeight/Width we shall not recalulating the minHeight. This seams to not have any negative visual impacts uppon testing.


carldea (approved):
Looks good. Thank you so much Fredrick @vollbrecht-work !
We will test the PR, before merging. And make the final release of Komet.
Great working with you and others!

dukke:
@rbradley813 @carldea this PR does not fix the ticket it is referring to... closing the PR.

